PR #12, opened 2018-04-12
Issue #11 - Make strftime path expansions in config skippable

### LINKED ISSUE #11: Config strftime expansions are always applied to paths
Config file loading attempts to expand out all `strftime` format strings in paths. There's no way to turn this off at the moment for a given string/path. 

The PCap rotation being used by #10 requires that a format string be passed to `open()` with `strftime` format strings so the log can be rotated. Currently this is using `command.history.filename` as the value but any` strftime` format strings in the name are expanded out on load. This results in all the logs being rotated to the same file name ...

We need to add a way to mark a string as not requiring the `strftime` expansion in the config file.

MJJoyce:
Update replaceVariable strftime expansion to ignore strings that end
with \`raw\`. Any string with \`raw\` at the end will be returned without
the \`raw\` tag and without further modification.

Resolve #11 

Note, this is a super hacky and ugly way of dealing with this. I can't help but think we need to come up with something a big more elegant than this ...


jordanpadams:
@MJJoyce maybe we update cfg.py instead to store both the raw and the strftime, and then programmatically you can grab either? this would be similar(ish) to how we handle raw versus dntoeu telemetry. thoughts?

PR #13, opened 2018-04-12
Issue #10 - Add optional command history file rotation to CmdAPI

### LINKED ISSUE #10: Add log rotation support to CmdAPI Command History file
Add log rotation support to the command history pcap writer in CmdAPI. This should be able to use the built in PCap rotation support.

MJJoyce:
Add config options to CmdAPI so command history writing be customized
with rotation based upon a number of factors. The CmdAPI object now
keeps the command history file open through its lifecycle instead of
opening it each time a command is sent. We may need to consider calling
`flush()` and `os.fsync()` to guarantee everything is written as
expected.

Added configuration:

    command.history.rollover.enable
        Defaults to True. Toggles whether command history files should
        be rotated or not.

    command.history.rollover.nbytes
        Defaults to None. The number of bytes written to the PCap file
        that should trigger a rotation.

    command.history.rollover.npackets
        Defaults to None. The number of packets written to the PCap file
        that should trigger a rotation

    command.history.rollover.nseconds
        Defaults to None. If both `nbytes` and `npackets` are None,
        defaults to 86400 (1 day). The number of seconds since the first
        packet is written to the PCap that should trigger a rotation.

Update example config so that the command history path includes strftime
format strings for proper log rotation.

Resolve #10


MJJoyce:
Note, this will impact GUI handling of command history files. Don't merge until a solution is added for that as well.

PR #22, opened 2018-04-16
Issue #16 - Make AIT-Core docs public on ReadTheDocs

### LINKED ISSUE #16: Get Core docs built and publicly viewable

lorsposto:
Resolves #16
- Adds readthedocs.yml
- Updates Sphinx version to 1.7.2
- Adds BLISS_CONFIG setting to conf.py if doesn't exist for RTD build purposes
- Remove project mentions from index

Caveats:
- Search doesn't work

Preview docs of this branch available at http://ait-core.readthedocs.io/en/issue-16/index.html. A spot check looked about on-par with the original docs.

Will need to add a webhook to repo settings for integration.


MJJoyce:
+1 from me. End result looks good! Can you squash the commits down so the history is a bit more coherent before merging?

Regarding the search issue, please open a follow up ticket so we can track that. I noticed the following in the console when loading the page. Don't know if it's relevant or not.
screen shot 2018-04-17 at 07 44 37

PR #86, opened 2018-06-06
Issue #85 - Expand core.db backend support

### LINKED ISSUE #81: TIME40 dtype doesn't take up the proper number of bytes
The TIME40 type definition is based around a `MSB_U32` primitive type. As such, the size of TIME40 is set as 4 bytes / 32 bits instead of proper size of 5 bytes / 40 bits.

### LINKED ISSUE #85: Expand core.db backend support
It would be good to have a more generic DB backend abstraction than what is currently in `core.db` and InfluxDB is a fairly popular time series database that we've considered adding support for through conversations with users. I propose that we add support for an InfluxDB backend much like we currently have for SQLite and use that as a driver for generalizing the current SQLite abstraction into something more generic while still keeping it lightweight and simple. I suggest keeping the interface simple and largely the way it currently is but generalizing it into an abstract base class with inherited drivers for SQLite and InfluxDB.

Something along the lines of the following public methods would cover most of the use cases that the toolkit ought to cover out of the box.
- connect: Connect to a database instance via the backend driver. Should probably take some config options for host/port/username/password, etc
- create: Create "tables" (or whatever the particular backend calls them) for each packet type
- insert: Given a packet, insert it it's data into the database
- query: Take a string in the query language that the database expects and return results
- close: Close the connection with the database and handle any cleanup

MJJoyce:
Add the GenericBackend abstract base class in core.db for defining the
expected interface for interacting with databases through AIT. Add
the SQLiteBackend sub-class which implements the previous core.db
backend functionality. Add the InfluxDBBackend sub-class which adds
support for InfluxDB to the toolkit.

Deprecate all previous core.db functions. Users should instead use the
SQLiteBackend class if they require the previous functionality. All
previous core.db functions will be removed in a future release.

Add tests for the new core.db backend classes

Update ait-tlm-db-insert to use the new backend classes. It also expects
to receive data in PCAPs instead of generic binary blobs in a file since
PCAPs are the project baseline for data storage.

Resolve #81 
Resolve #85

### DIFF
--- a/ait/core/bin/ait_tlm_db_insert.py
+++ b/ait/core/bin/ait_tlm_db_insert.py
@@ -15,7 +15,7 @@
 # information to foreign countries or providing access to foreign persons.
 
 """
-Inserts telemetry into a database from one or more files.
+Inserts telemetry into a database from one or more PCAP files.
 """
 
 
@@ -23,10 +23,11 @@ import argparse
 import os
 import sys
 import socket
+import struct
 import time
 
 import ait
-from ait.core import db, log, tlm
+from ait.core import db, log, tlm, pcap
 
 
 def main():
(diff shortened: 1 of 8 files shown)


MJJoyce:
Note, I ran into #81 while testing the new database backends and included the fix in the branch. I've tested both of the backends with basic data inserts and retrievals locally. I'd like to test the SQLite driver a bit more with Time/Cmd/EVR fields more to make sure the logic for column types is correct. It wasn't handling this properly in the previous abstractions and I'm not 100% certain it's perfect now Specifically, Time8 is going to be listed as a float instead of an integer with the current implementation. I don't think it'll matter but I would like to double check first. The docs for the backends need fleshed out and a main database page needs added to the docs since nothing currently exists for the old implementation.

- [ ] Double check that the typing of columns for complex datatypes works as expected with the SQLite driver
- [ ] Flesh out docstrings for the database backends. Specifically need to highlight how some functionality differs in the specific implementations vs the ABC expectations
- [ ] Add a databases page to the docs build. At the moment we don't highlight anywhere that this functionality exists for users.

MJJoyce:
Alright, good to merge this now. Tested with SQLite and updated the docs as needed

PR #98, opened 2018-07-01
Issue #97: add configurable subject and from address

### LINKED ISSUE #97: Allow configurable Subject and From for ait.notify notifications

jordanpadams:
* Subject defaults to 'AIT Notification'.
* From defaults to 'ait-notify@my.smtp.server'

This update also includes a bug fix for s.sendmail(). First argument is
supposed to be from_addr [1]:

  SMTP.sendmail(from_addr, to_addrs, msg[, mail_options, rcpt_options])

[1] https://docs.python.org/2/library/smtplib.html#smtplib.SMTP.sendmail.

Testing in past must have used SMTP server that did not check to verify
from_addr was an email format (e.g. ait@gmail.com).

Another note from testing is the from_addr MUST have a valid @server in
order to work for text messaging. At least while testing with AT&T, it appears
they perform a check to verify it is a legitimate domain prior to sending.
Interestingly enough, it does not bounce the message with an error, it
simply drops it to the floor.

Resolves #97

### DIFF
--- a/ait/core/notify.py
+++ b/ait/core/notify.py
@@ -66,15 +66,20 @@ def _send_email(message, recipients):
         log.error('Email SMTP connection parameter error. Please check config.')
         return
 
+    subject = ait.config.get('notifications.smtp.subject', 'AIT Notification')
+
+    # From address must have a valid @server, otherwise texts will not works
+    fromaddr = ait.config.get('notifications.smtp.from', 'ait-notify@%s' % server)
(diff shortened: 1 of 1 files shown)


jordanpadams:
@NASA-AMMOS/ait-project-management-committee can we get this looked at, approved, and potentially get a point build out ASAP for ECOSTRESS?

PR #113, opened 2018-09-04
Issue #112 - Expand EVR Format Message Handling

### LINKED ISSUE #112: Expand EVR Format Message
The current EVR format message handling does an ok job of processing format strings and displaying them. However, the current functionality doesn't result in the expected output in all cases and we don't provide any support for size format strings. It would be useful to expand on the functionality for this use case.

MJJoyce:
EVR message format string processing has been updated to better handle
type and size format information. Data conversion has been switched over
to the dtype.PrimitiveType class so that processing of binary values
into ints/floats returns expected values.

Resolve #112

### DIFF
--- a/ait/core/evr.py
+++ b/ait/core/evr.py
@@ -24,8 +24,7 @@ import pkg_resources
 import re
 import yaml
 
-import ait
-from ait.core import json, log, util
+from ait.core import dtype, json, log, util
 
 
 class EVRDict(dict):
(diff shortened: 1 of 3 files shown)


MJJoyce:
ooopss, didn't notice an import issue in my local install. I'll get the PR fixed and pushed shortly

MJJoyce:
Looks like we had a circular import problem of sorts with the new changes (`evr.py` imported `dtype.py` and vice versa). I've moved the `evr.py` import in `dtype.py` to just before a check for fetching the EVR dictionary to resolve this

ait/core/evr.py
```diff
@@ -127,43 +128,74 @@ def format_message(self, evr_hist_data):
...
+                f_size_char = f[:-1]
+
+            fsize = size_formatter_info[f_size_char.lower()]
 
             try:
-                # Normal data chunking is the current byte index + the size
-                # of the relevant data type for the formatter
-                if format_size > 0:
-                    end_index = cur_byte_index + format_size
+                if f != 's':
```
MJJoyce: It probably makes more sense to check if `f_type == 's'` here since we're using `f` to seed that value. It shouldn't change anything if left as is, but probably clearer updated.

PR #114, opened 2018-09-05
Issue #83 - Baseline example configs across repos

### LINKED ISSUE #83: Baseline example configs across repos
We should update all of the example configs that we include with the repositories so they're the same across all the repositories.

aywaldron:
Changes are as follows:
*  Added missing dsn configs to core `config.yaml` and `tlm.yaml`
*  Adding files present in `ait-dsn/config` but missing from`ait-core/config`
*  Removed extra data directory layer so `data/config` -> `config/`
*  Removing references to removed `data/` directory layer.
*  Removed limits directory layer within config directory so `config/limits/` -> `config`
*  Removing references to removed `limits/` directory layer.

I also added `build/lib/*` and `*.DS_Store` to .gitignore

Resolves #83


config/mib/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
```
MJJoyce: Let's drop this and leave it in the DSN repo

config/sle_tlm.yaml
```diff
@@ -0,0 +1,194 @@
+- !Packet
```
MJJoyce: Let's drop this and leave it in the DSN repo

config/tlm.yaml
```diff
@@ -86,3 +86,5 @@
       value: 0x89ABCDEF
 
 - !include ccsds_header.yaml
+- !include sle_tlm.yaml
```
MJJoyce: Let's drop this since the file will only be in the DSN repo after the above changes

PR #126, opened 2018-10-19
Issue #7 - Add support for derived telemetry values

### LINKED ISSUE #7: Add support for derived telemetry values
EGSE users have specified that they would like to be able to define derived telemetry points, especially when scripting instrument functionality with `bliss.core.api`, so they can track various stats that might not be returned in telemetry. For instance, they might track the number of times that a sensor's power state has been toggled on/off.

Whether this would be injected into telemetry packets in the `api` or implemented and accessed some other way is still TBD

aywaldron:
First stab at derived telemetry support. 

Resolves #7

### DIFF
--- a/ait/core/test/test_tlm.py
+++ b/ait/core/test/test_tlm.py
@@ -19,6 +19,7 @@ import csv
 import struct
 
 import nose
+from nose.tools import *
 
 from ait.core import tlm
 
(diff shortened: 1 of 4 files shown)


aywaldron:
@MJJoyce Just pushed a commit that should address this.

ait/core/test/test_tlm.py
```diff
@@ -118,6 +119,85 @@ def test_fld_defn_notitle(self):
         os.remove(test_yaml_test2)
 
 
+class TestDerivationDefinition(object):
+    test_yaml_deriv1 = '/tmp/test_deriv1.yaml'
+
+    yaml_docs_deriv1 = (
+        '- !Packet\n'
+        '  name: OCO3_1553_EHS\n'
```
MJJoyce: Let's make a follow up ticket and make some changes to the names for consistency

PR #136, opened 2019-01-11
Issue #99 - Add documentation about default command encoding format

### LINKED ISSUE #99: Command encoding is ISS specific
The current command encoding is ISS specific. From the `encode()` docstring

```
 def encode(self, pad=106):
        """Encodes this AIT command to binary.

        If pad is specified, it indicates the maximum size of the encoded
        command in bytes.  If the encoded command is less than pad, the
        remaining bytes are set to zero.

        Commands sent to ISS payloads over 1553 are limited to 64 words
        (128 bytes) with 11 words (22 bytes) of CCSDS overhead (SSP
        52050J, Section 3.2.3.4).  This leaves 53 words (106 bytes) for
        the command itself.
```

We don't necessarily need to change this default since we support [extensions](https://ait-core.readthedocs.io/en/latest/extensions.html). However, we should make sure the default is documented clearly.

MJJoyce:
Resolve #99

### DIFF
--- a/doc/source/command_intro.rst
+++ b/doc/source/command_intro.rst
@@ -68,7 +68,7 @@ All the valid parameters and attributes that you can have in your command dictio
 AIT also provides a command line utility for verifying that your command dictionary configuration is valid given that you have a defined schema file. If you pass the ``--cmd`` or ``-c`` flag to ``ait-yaml-validate`` it will check this for you.
 
 .. code-block:: bash
-    
+
     $ ait-yaml-validate --cmd
     016-07-27T09:36:21.408 | INFO     | Validation: SUCCESS: ...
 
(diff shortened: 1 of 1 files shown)


doc/source/command_intro.rst
```diff
@@ -97,6 +97,17 @@ Given a binary blob, you can also decode into a command.
...
+----------------------------------
+
+By default AIT encodes commands into 64 word (128 byte) structure with 11 words (22 bytes) of CCSDS overhead. This results in a 53 words (106 bytes) of usable commanding space. All commands are padded out to 106 bytes.
+
++------------------+--------------------+-----------------------------+-----------+-----+
+| Opcode (2 bytes) | Args Size (1 byte) | Arg 1 Encoded (size varies) | Arg 2 ... | ... |
++------------------+--------------------+-----------------------------+-----------+-----+
+
+If you need a different format of command encoding the `extensions <extensions.html>`__ functionality will allow you to override the default behavior.
+
```
aywaldron: The extensions link is broken for me.

aywaldron: Works for me now!

Issue #75, opened 2018-05-23
Create script for generating dummy telemetry data for tests


MJJoyce:
Per @MJJoyce 

Having a script for generating dummy telemetry data with marching values, data within specific ranges, etc. would be useful when trying to test new functionality.

---

Per @bbornstein  

Yes, that would be great.  Given a `PacketDefinition`, it's straightforward:

```python
defn = tlm.getDefaultDict()[packetName]
data = bytearray(xrange(defn.nbytes))
pkt  = tlm.Packet(defn, data)
```

This should really be a method on `PacketDefinition`, e.g.:

```python
def simulate(fill=None):
    size   = self.nbytes
    values = xrange(size) if fill is None else ((fill,) * size)
    return Packet(self, bytearray(data))
```

(I chose `simulate()` as it seems somehow analogous to `validate()`.)   Any script could then easily simulate packets:

```python
defn = tlm.getDefaultDict()[packetName]
pkt  = defn.simulate()
...
```                    

---

Per @MJJoyce 
                       
+1, I like the `simulate()` choice as well

PR #140, opened 2019-02-26
GUI Issue #75 - Backend Reimagining

aywaldron:
Resolves https://github.com/NASA-AMMOS/AIT-GUI/issues/75


ait/core/server/handler.py
```diff
@@ -0,0 +1,106 @@
...
+        self.input_type = input_type
+        self.output_type = output_type
+
+        for key, value in kwargs.items():
+            setattr(self, key, value)
+
+    def __repr__(self):
+        return '<handler.%s>' % (self.__class__.__name__)
+
+    def validate_input(self, input_data):
```
MJJoyce: I'm still uncertain of whether we need to care about validating stuff. Could you elaborate on why you think it's a good choice? I'm probably overlooking something and would like to hear other perspectives.

ait/core/server/handler.py
```diff
@@ -0,0 +1,106 @@
...
+        """
+        Checks that the input data to the handler matches the expected
+        input type. If input data is of the correct type, calls the
+        handle method. If input data is not the correct type, raises a
+        ValueError.
+        """
+        try:
+            checked_input = self.validate_input(input_data)
+        except Exception as e:
+            raise(ValueError('Input is not of valid type: ', e))
```
MJJoyce: Are we catching this `ValueError` when we pass data through handlers in our streams? We should make sure that's resilient to issues (regardless of the `validate_input`)

ait/core/server/handler.py
```diff
@@ -0,0 +1,106 @@
...
+            raise ValueError(msg)
+
+        self._pkt_defn = tlm_dict[self.packet]
+
+    def handle(self, input_data):
+        # this is being published to ZMQ - needs to work
+        return (self._pkt_defn.uid, input_data)
+
+
+class CcsdsPacketHandler(Handler):
```
MJJoyce: We need to update this so we're handling  to match the current Apid handle for CCSDS stuff.

ait/core/server/handler.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def __init__(self, input_type, output_type=None):
+        super(CcsdsPacketHandler, self).__init__(input_type, output_type)
+
+    def handle(self, input_data):
+        return input_data + 5
+
+
+class TmTransFrameDecodeHandler(Handler):
```
MJJoyce: Is this based off of some existing functionality? Should up this (or remove it if we don't need it)

ait/core/server/plugin.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+class Plugin(ZMQInputClient):
+    """
+    This is the parent class for all plugins. All plugins must implement
+    their own process method which is called when a message is received.
+    """
+
+    def __init__(self, inputs, outputs, zmq_args={}, **kwargs):
+        self.type = 'Plugin'
```
MJJoyce: What are we doing with `type`? Do we need this?

ait/core/server/plugin.py
```diff
@@ -0,0 +1,138 @@
...
+
+import gevent
+import gevent.monkey; gevent.monkey.patch_all()
+
+import ait.core
+from ait.core import  db, limits, log, notify, tlm
+from client import ZMQInputClient
+
+
+class Plugin(ZMQInputClient):
```
MJJoyce: I think this should also be an abstract base class. See `Handler` for additional notes.

ait/core/server/plugin.py
```diff
@@ -0,0 +1,138 @@
...
+
+        try:
+            mod, cls = self.datastore.rsplit('.', 1)
+            dbconn = getattr(importlib.import_module(mod), cls)()
+            dbconn.connect(**kwargs)
+        except ImportError:
+            log.error("Could not import specified datastore {}".format(self.datastore))
+            return
+        except Exception as e:
+            log.error("Unable to connect to InfluxDB backend. Disabling data archive.\n{}".format(e))
```
MJJoyce: This error message should be backend agnostic

ait/core/server/plugin.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+class DataArchive(Plugin):
+
+    def __init__(self, inputs, outputs, datastore='ait.core.db.InfluxDBBackend', **kwargs):
+        super(DataArchive, self).__init__(inputs, outputs, **kwargs)
+
+        self.datastore = datastore
+
+    def process(self, input_data, topic=None, **kwargs):
```
MJJoyce: A log of this should get pulled out of process and moved to the `__init__`. Parsing the telemetry dictionary into a helpful format, setting up a database connection, etc. shouldn't be happening every time we get a packet and process it.

ait/core/server/plugin.py
```diff
@@ -0,0 +1,138 @@
...
+        except ImportError:
+            log.error("Could not import specified datastore {}".format(self.datastore))
+            return
+        except Exception as e:
+            log.error("Unable to connect to InfluxDB backend. Disabling data archive.\n{}".format(e))
+            return
+
+        try:
+            log.info('Starting telemetry data archiving')
+            split = re.split(r'\((\d),(\'.*\')\)', input_data)
```
MJJoyce: I'd rather we not split a string with regex if we can help it.

ait/core/server/plugin.py
```diff
@@ -0,0 +1,138 @@
...
+            dbconn.connect(**kwargs)
+        except ImportError:
+            log.error("Could not import specified datastore {}".format(self.datastore))
+            return
+        except Exception as e:
+            log.error("Unable to connect to InfluxDB backend. Disabling data archive.\n{}".format(e))
+            return
+
+        try:
+            log.info('Starting telemetry data archiving')
```
MJJoyce: This prints every time we get a packet. Should get moved

ait/core/server/plugin.py
```diff
@@ -0,0 +1,138 @@
...
+            split = re.split(r'\((\d),(\'.*\')\)', input_data)
+            uid, pkt = int(split[1]), split[2]
+            defn = packet_dict[uid]
+            decoded = tlm.Packet(defn, data=bytearray(pkt))
+            dbconn.insert(decoded, **kwargs)
+
+            gevent.sleep(0)
```
MJJoyce: We're already calling gevent.sleep as part of the Plugin's run loop. Are we getting something from doing this here as well?

ait/core/server/server.py
```diff
@@ -0,0 +1,255 @@
...
+
+        gevent.joinall(self.greenlets)
+
+    def _load_streams(self):
+        """
+        Reads, parses and creates streams specified in config.yaml.
+        """
+        common_err_msg = 'No valid {} telemetry stream configurations found. '
+        specific_err_msg = {'inbound': 'No telemetry will be received (or displayed).',
+                            'outbound': 'No telemetry will be published.'}
```
MJJoyce: Nit-picky, but we probably want something other than `telemetry` specified here. No `data` or something similar.

ait/core/server/server.py
```diff
@@ -0,0 +1,255 @@
...
+                                                                        value))
+        if not self.inbound_streams and not self.servers:
+            log.warn(err_msgs['inbound'])
+
+        if not self.outbound_streams:
+            log.warn(err_msgs['outbound'])
+
+    def _create_stream(self, config, stream_type):
+        """
+        Creates a stream from its config.
```
MJJoyce: Let's make sure we get all the function docstrings updated to this format. Although a little more whitespace between the categories of stuff and a more elaborate description of attributes would also be nice =)

ait/core/server/server.py
```diff
@@ -0,0 +1,255 @@
...
+            stream_type:  either 'inbound' or 'outbound'
+        Returns:
+            stream:       a Stream
+        Raises:
+            ValueError:   if any of the required config values are missing
+        """
+        if stream_type not in ['inbound', 'outbound']:
+            raise ValueError('Stream type must be \'inbound\' or \'outbound\'.')
+
+        if config is None:
```
MJJoyce: Is this ever going to fail? Let's default the config parameter to None in the signature if we want to make this check I think.

ait/core/server/server.py
```diff
@@ -0,0 +1,255 @@
...
+
+        Params:
+            config:       stream configuration as read by ait.config
+            stream_type:  either 'inbound' or 'outbound'
+        Returns:
+            stream:       a Stream
+        Raises:
+            ValueError:   if any of the required config values are missing
+        """
+        if stream_type not in ['inbound', 'outbound']:
```
MJJoyce: Is this check necessary? This is a "private" function and it can only ever get one of these two values. Super nit-picky and not the end of the world, but a bit of a code smell to me.

ait/core/server/server.py
```diff
@@ -0,0 +1,255 @@
...
+            raise ValueError('No stream config to create stream from.')
+
+        name = config.get('name', None)
+        if name is None:
+            raise(cfg.AitConfigMissing(stream_type + ' stream name'))
+        if name in [x.name for x in (self.outbound_streams +
+                                     self.inbound_streams +
+                                     self.servers +
+                                     self.plugins)]:
+            raise ValueError('Stream name already exists. Please rename.')
```
MJJoyce: A more descriptive error message would be good here. Something that mentions the duplicate name. E.g., `'Duplicate stream name "{name}" encountered. Stream names must be unique.'`

ait/core/server/server.py
```diff
@@ -0,0 +1,255 @@
...
+
+        stream_input = config.get('input', None)
+        if stream_input is None and stream_type is 'inbound':
+            raise(cfg.AitConfigMissing(stream_type + ' stream input'))
```
MJJoyce: Might need a more descriptive message here

ait/core/server/server.py
```diff
@@ -0,0 +1,255 @@
...
+
+        stream_handlers = [ ]
+        if 'handlers' in config:
```
MJJoyce: We've got a lot going on in this function and it looks like there's good places (here / below, but also others) to split code out into functions that helps us keep stuff readable / testable.

ait/core/server/server.py
```diff
@@ -0,0 +1,255 @@
...
+        if plugins is None:
+            log.warn('No plugins specified in config.')
+        else:
+            for index, p in enumerate(plugins):
+                try:
+                    plugin = self._create_plugin(p['plugin'])
+                    self.plugins.append(plugin)
+                    log.info('Added plugin {}'.format(plugin))
+
+                except Exception:
```
MJJoyce: Any general `Exception` catch needs to be updated to be more explicit and catch whatever exception(s) we're worried about. There's always exceptions to the rule, of course, but I don't think the examples in here are those.

ait/core/server/server.py
```diff
@@ -0,0 +1,255 @@
...
+                    exc_type, value, tb = sys.exc_info()
+                    log.error('{} creating plugin {}: {}'.format(exc_type,
+                                                                 index,
+                                                                 value))
+            if not self.plugins:
```
MJJoyce: This check seems a bit off to me. Seems like a double check with the previous if/else block that isn't necessary. Maybe something that got left in and needs cleaned out?

ait/core/server/server.py
```diff
@@ -0,0 +1,255 @@
...
+            plugin:       a Plugin
+        Raises:
+            ValueError:   if any of the required config values are missing
+        """
+        if config is None:
+            raise ValueError('No plugin config to create plugin from.')
+
+        name = config.get('name', None)
+        if name is None:
+            raise(cfg.AitConfigMissing('plugin name'))
```
MJJoyce: Better message here?

ait/core/server/server.py
```diff
@@ -0,0 +1,255 @@
...
+
+        # TODO I don't think we actually care about this being unique? Left over from
+        # previous conversations about stuff?
+        module_name = name.rsplit('.', 1)[0]
+        class_name = name.rsplit('.', 1)[-1]
+        if class_name in [x.name for x in (self.outbound_streams +
```
MJJoyce: We might need to rethink this "only one plugin of a given name" rule. I think it makes sense to have a unique `id` of some sort associated with a plugin but I don't think that should necessarily be the class name. I can think of some weird, but not totally impossible, use case for having multiples of the same plugin setup (especially the data archiving stuff). Probably fair to follow up on this is a different ticket though.

ait/core/server/stream.py
```diff
@@ -0,0 +1,90 @@
...
+        Invokes each handler in sequence.
+        Publishes final output data.
+        """
+        for handler in self.handlers:
+            output = handler.execute_handler(input_data)
+            input_data = output
+
+        self.publish(input_data)
+
+    def valid_workflow(self):
```
MJJoyce: Similarly to the `validate_input` comment, do we care about this? I'm not convinced and would like to hear opinions.

ait/core/server/handler.py
aywaldron: +1 need to catch and log this error at the stream level

ait/core/server/plugins/ait_gui_plugin.py
```diff
@@ -0,0 +1,1157 @@
+import gevent
```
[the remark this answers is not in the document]
aywaldron: ^ before merging

ait/core/server/handler.py
aywaldron: +1 should remove this

ait/core/server/plugin.py
aywaldron: Update usage to just call type() on class, removing need for type attribute

ait/core/server/plugin.py
aywaldron: (use `split = input_data[1:-1].split(',', 1)` as in TelemetryLimitMonitorPlugin below

ait/core/server/server.py
aywaldron: Split `create_stream` into 2 separate functions for inbound and outbound stream types

ait/core/server/server.py
aywaldron: ^ will change later

ait/core/server/server.py
aywaldron: Add stream name to message

ait/core/server/server.py
aywaldron: not possible here since plugin name not available

ait/core/server/server.py
aywaldron: This is to make sure the server doesn't crash on creation of plugins no matter what is raised

ait/core/server/server.py
aywaldron: This is for the case where plugins are specified in the config, but none of those configs are valid

ait/core/server/handler.py
aywaldron: (and this should not be a generic exception catch that then raises a ValueError)

ait/core/server/handler.py
aywaldron: ^ Don't actually need CCSDS packet handler at the moment. might need in future. remove for now.

doc/source/server_architecture.rst
```diff
@@ -0,0 +1,86 @@
+AIT Server Introduction
+========================
+
+The core of AIT is a "server", :mod:`ait.core.server`, which manages the processing of telemetry, logging, and command messages. The server manages relationships between the following components: 
+
+* **inbound streams**   - inbound streams manage incoming data being recieved by AIT, for example telemtry messages coming into a port.
```
MJJoyce: There's a few typos here. Updated along with slight format change:
```
Inbound streams manage incoming data being received by AIT. For example, telemetry messages coming into a port.
```

I think the same format change for outbound streams make sense as well. Similarly, capitalize all the list items helps clean it up a bit IMO.

doc/source/server_architecture.rst
```diff
@@ -0,0 +1,86 @@
...
+Streams, Plugins and Handlers
+-----------------------------
+
+Data will generally flow through the components as follows:
+
+-> Inbound Stream -> (Handlers) -> Plugin -> Outbound Stream -> (Handlers) ->
+
+* Plugins can have any number of inbound streams as inputs and any number of outbound streams as outputs. 
+* Inbound streams can have either integer ports or other inbound stream names as inputs.
+* Outbound streams can have either plugin or inbound stream names as inputs. They can output to other outbound streams or to an integer port.
```
MJJoyce: I think this first sentence should say **outbound** stream not inbound

doc/source/server_architecture.rst
```diff
@@ -0,0 +1,86 @@
...
+The core of AIT is a "server", :mod:`ait.core.server`, which manages the processing of telemetry, logging, and command messages. The server manages relationships between the following components: 
+
+* **inbound streams**   - inbound streams manage incoming data being recieved by AIT, for example telemtry messages coming into a port.
+* **outbound streams**   - outbound streams manage outgoing data being published by AIT, for example command messages being published to a port.
+* **handlers**   - handlers do the processing of data for inbound and outbound streams. When a message is received on an inbound stream for instance, a handler can handle the processing that needs to be done on the message before it is passed off to outbound streams or plugins.
+* **plugins**       - plugins provide functionality that consume data from inbound streams, do something with it, and provide data to outbound streams. The AIT GUI is a plugin.
+
+The AIT Server uses the messaging library ZeroMQ to manage the passing of data between streams and plugins.
+
+Streams, Plugins and Handlers
```
MJJoyce: I think adding an individual section on each of the the components makes sense here. Have the main overarching section and then add subsections for Streams, Plugins, and handlers. You could then elaborate on what each of those does in detail, link to the API docs for each, explain creating custom ones, etc. while keeping everything nicely organized.

doc/source/server_architecture.rst
```diff
@@ -0,0 +1,86 @@
...
+Configuring the server
+----------------------
+
+AIT uses `**config.yaml** <https://ait-core.readthedocs.io/en/master/configuration_intro.html>` to load configuration data for the server.
+
+Here is an example of the **server** portion of **config.yaml**:
+
+.. code-block:: none
+
+    server:
```
MJJoyce: This config snipped needs updated. It has the "old" plugin name import style and the outbound stream section has the old Plugin input stuff that got changed.

doc/source/server_architecture.rst
```diff
@@ -0,0 +1,86 @@
...
+                    - name: ait.server.handlers.ait_packet_handler
+                      packet: 1553_HS_Packet
+
+        outbound-streams:
+            - stream:
+                name: command_stream
+                input: AitGuiPlugin
+                output: 3075
+
+Plugins
```
MJJoyce: Here's those sections I as talking about. I think including these configuration subsections with the stuff mentioned above would make sense

ait/core/server/handler.py
aywaldron: I agree this is unnecessary. Will remove.

ait/core/server/handler.py
aywaldron: Removed `vaildate_input` functionality

ait/core/server/plugin.py
aywaldron: nope, will remove

doc/source/server_architecture.rst
aywaldron: Outbound streams can have either inbound or outbound streams as inputs. A plugin is not required.

ait/core/server/stream.py
aywaldron: This would be helpful to the user in setting up their handlers, so they would know from the beginning if their handlers weren't able to be executed in sequence. It is an optional functionality which only checks compatibility if output and input type of next handler are both specified.

aywaldron:
Got an offline +1 from @MJJoyce to merge.

Issue #157, opened 2019-06-06
Add CCSDSHandler class


seanlu99:
The CCSDSHandler will allow us to accept multiple packet types on a single stream and have them be processed.

PR #158, opened 2019-06-06
Issue #158- Add CCSDSHandler class

seanlu99:


ait/core/server/handler.py
```diff
@@ -84,3 +85,51 @@ def handle(self, input_data):
...
+class CCSDSPacketHandler(Handler):
+
+    def __init__(self, input_type=None, output_type=None, **kwargs):
+        """
+        Params:
+            input_type:   (optional) Specifies expected input type, used to
+                                     validate handler workflow. Defaults to None.
+            output_type:  (optional) Specifies expected output type, used to
+                                     validate handler workflow. Defaults to None
+            **kwargs:     (required) APID values: packet name pairs
```
MJJoyce: Instead of specifying `**kwargs` here it would be good to specify what keyword arguments are required / optional.

For example,
```
packet_types: (required) APID values : packet name pairs
packet_secondary_header_length: (optional) ...
```

Otherwise someone reading the docs has to go look in the code to figure out what the name for the keyword arguments you're describing are.

ait/core/server/handler.py
```diff
@@ -84,3 +85,51 @@ def handle(self, input_data):
             tuple of packet UID and message received by stream
         """
         return pickle.dumps((self._pkt_defn.uid, input_data), 2)
+
+
+class CCSDSPacketHandler(Handler):
```
MJJoyce: Please add a docstring to the class describing what it does at a reasonable level of detail so the generated documentation provides helpful docs for people.

ait/core/server/handler.py
```diff
@@ -84,3 +85,51 @@ def handle(self, input_data):
...
+            input_type:   (optional) Specifies expected input type, used to
+                                     validate handler workflow. Defaults to None.
+            output_type:  (optional) Specifies expected output type, used to
+                                     validate handler workflow. Defaults to None
+            **kwargs:     (required) APID values: packet name pairs
+                          (optional) Secondary header length
+        """
+        super(CCSDSPacketHandler, self).__init__(input_type, output_type)
+        self.packet_types = kwargs['packet_types']
+        ait.core.log.info(self.packet_types)
```
MJJoyce: I don't think we need to leave this log message here? Was this accidentally left in from testing?

ait/core/server/handler.py
```diff
@@ -84,3 +85,51 @@ def handle(self, input_data):
...
+        self.packet_secondary_header_length = kwargs.get('packet_secondary_header_length', 0)
+        ait.core.log.info(self.packet_secondary_header_length)
```
MJJoyce: Same as above

ait/core/server/handler.py
```diff
@@ -84,3 +85,51 @@ def handle(self, input_data):
...
+        """
+        Params:
+            packet:    CCSDS packet
+        Returns:
+            tuple of packet UID and packet data field
+        Raises:
+            ValueError:   If packet is specified but not present in default tlm dict.
+        """
+        packet = bytearray(input_data)
+        packet_apid = int(binascii.hexlify(packet[6:8]), 16) & 0x7FF
```
MJJoyce: Since you're bitwise "and-ing" two bytes here I'd like to see `0x7FF` listed with the "correct" number of values. I realize this still works as expected but `0x07FF` is extra clear.

ait/core/server/handler.py
```diff
@@ -84,3 +85,51 @@ def handle(self, input_data):
...
+        if packet_apid not in self.packet_types:
+            msg = 'CCSDSPacketHandler: Packet APID {} not present in config'.format(packet_apid)
+            msg += ' Available packet APIDs are {}'.format(self.packet_types.keys())
+            raise ValueError(msg)
```
MJJoyce: Let's print an error and carry on if we get an invalid APID. We shouldn't pull down the handler for one bad APID (since it's far from impossible that a bit will get flipped somewhere, etc.).

ait/core/server/handler.py
```diff
@@ -84,3 +85,51 @@ def handle(self, input_data):
...
+        packet_name = self.packet_types[packet_apid]
+        tlm_dict = tlm.getDefaultDict()
+        if packet_name not in tlm_dict:
+            msg = 'CCSDSPacketHandler: Packet name {} not present in telemetry dictionary'.format(packet_name)
+            msg += ' Available packet types are {}'.format(tlm_dict.keys())
+            raise ValueError(msg)
```
MJJoyce: I'm more okay with this check raising a value error but this packet name check should be done in `__init__` before the handler ever starts looking at data. In other words, part of the initial "boot up" in `__init__` should be to look at the APID packet mappings and make sure the packet names listed are valid. Then the user will see an obvious error when they boot up their server and not be surprised when they're handling data (potentially much) later.

config/config.yaml
```diff
@@ -127,8 +127,19 @@ default:
                 handlers:
```
MJJoyce: Let's drop the config updates from the PR. We want to leave the defaults alone unless we explicitly need to update it.

.idea/workspace.xml
```diff
@@ -0,0 +1,217 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
MJJoyce: Drop this. Should probably update your local .gitignore to avoid this file in the future =D

MJJoyce:
@seanlu99, Thanks so much for opening a ticket and making this pull request. I had a few comments that I think could help clean it up a bit. If you have questions let me know.

Also, please checkout our contribution guide and make updates so your PR follows those guidelines. Specifically, 
https://github.com/NASA-AMMOS/AIT-Core/wiki/Contributing#commit-messages
https://github.com/NASA-AMMOS/AIT-Core/wiki/Contributing#pull-requests-and-feature-branches (regarding the closing of tickets via message in the PR)

MJJoyce:
@seanlu99, I haven't looked through the other changes yet (will do shortly) but you deleted the entire `config.yaml` in your commit. Please update that!

aywaldron:
@seanlu99 I know this is closed already, but for the future: as far as I can tell the title should be Issue #157, not #158, (if this is the issue its referencing: https://github.com/NASA-AMMOS/AIT-Core/issues/157) and the description should include the text "Resolves #157" so that the issue would be closed automatically when the PR is merged. Commit messages should also contain "#157" so that commits are automatically linked to the issue ticket.

seanlu99:
@aywaldron My bad, will do next time

PR #161, opened 2019-06-10
Issue #159 - Stream handles None return from handlers

### LINKED ISSUE #159: Stream handler calls should handle None return from handlers
It's feasible that a handler will get input data that it can't process. With the Handler implementation this means we need to return `None` or something similar. Having the stream check for `None` and break the handler chain makes the most sense to me.

Current stream `process()`
https://github.com/NASA-AMMOS/AIT-Core/blob/master/ait/core/server/stream.py#L64
```
for handler in self.handlers:
    output = handler.handle(input_data)
    input_data = output

 self.publish(input_data)
```

Updating to something like the below would cover this situation
```
for handler in self.handlers:
    output = handler.handle(input_data)

    if output:
        input_data = output
    else:
        return

 self.publish(input_data)
```

seanlu99:
Resolves #159

### DIFF
--- a/ait/core/server/handler.py
+++ b/ait/core/server/handler.py
@@ -127,10 +127,11 @@ class CCSDSPacketHandler(Handler):
         """
 
         packet = bytearray(input_data)
-        if (len(packet) < 13):
+        if (len(packet) < 7):
             ait.core.log.info('CCSDSPacketHandler: Received packet length is less than minimum of 7 bytes.')
             return
-        packet_apid = int(binascii.hexlify(packet[6:8]), 16) & 0x07FF
+        packet_apid = int(binascii.hexlify(packet[0:2]), 16) & 0x07FF
+        ait.core.log.info(packet_apid)
         if packet_apid not in self.packet_types:
             msg = 'CCSDSPacketHandler: Packet APID {} not present in config.'.format(packet_apid)
             msg += ' Available packet APIDs are {}'.format(self.packet_types.keys())
@@ -140,11 +141,12 @@ class CCSDSPacketHandler(Handler):
         tlm_dict = tlm.getDefaultDict()
         packet_uid = tlm_dict[packet_name].uid
 
-        packet_data_length = int(binascii.hexlify(packet[10:12]), 16) + 1
-        if (len(packet) < 12 + packet_data_length):
+        packet_data_length = int(binascii.hexlify(packet[4:6]), 16) + 1
+        if (len(packet) < 6 + packet_data_length):
             ait.core.log.info('CCSDSPacketHandler: Packet data length is less than stated length in packet primary header.')
             return
         udf_length = packet_data_length - self.packet_secondary_header_length
-        udf_start = 12 + self.packet_secondary_header_length
+        udf_start = 6 + self.packet_secondary_header_length
         user_data_field = packet[udf_start:udf_start + udf_length]
+        ait.core.log.info(user_data_field[0])
         return pickle.dumps((packet_uid, user_data_field), 2)
--- a/ait/core/server/handler.py
+++ b/ait/core/server/handler.py
@@ -132,6 +132,7 @@ class CCSDSPacketHandler(Handler):
         packet_apid = int(binascii.hexlify(packet[6:8]), 16) & 0x07FF
         if packet_apid not in self.packet_types:
             msg = 'CCSDSPacketHandler: Packet APID {} not present in config.'.format(packet_apid)
+            msg += ' Available packet APIDs are {}'.format(self.packet_types.keys())
             ait.core.log.info(msg)
             return
         packet_name = self.packet_types[packet_apid]
--- a/ait/core/server/handler.py
+++ b/ait/core/server/handler.py
@@ -125,9 +125,9 @@ class CCSDSPacketHandler(Handler):
         Returns:
             tuple of packet UID and packet data field
         """
-
+        primary_header_length = 6
         packet = bytearray(input_data)
-        if (len(packet) < 7):
+        if (len(packet) < primary_header_length + 1):
             ait.core.log.info('CCSDSPacketHandler: Received packet length is less than minimum of 7 bytes.')
             return
         packet_apid = int(binascii.hexlify(packet[0:2]), 16) & 0x07FF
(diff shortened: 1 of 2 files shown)


ait/core/server/handler.py
```diff
@@ -72,8 +71,8 @@ def __init__(self, input_type=None, output_type=None, **kwargs):
         tlm_dict = tlm.getDefaultDict()
         if self.packet not in tlm_dict:
             msg = 'PacketHandler: Packet name {} not present in telemetry dictionary'.format(self.packet)
-            msg += ' Available packet types are {}'.format(tlm_dict.keys())
-            raise ValueError(msg)
```
MJJoyce: I think we should still be raising `ValueError` here. If the configured packet type isn't present in the config the handler can't do anything.

ait/core/server/handler.py
```diff
@@ -125,18 +123,24 @@ def handle(self, input_data):
         Returns:
             tuple of packet UID and packet data field
         """
+
         packet = bytearray(input_data)
+        if (len(packet) < 13):
+            ait.core.log.info('CCSDSPacketHandler: Insufficient packet length.')
```
MJJoyce: This is good but I'd like to see a more descriptive message. "Insufficient packet length" isn't wrong but something such as "Received packet is less than minimum of 12 bytes. Cannot process CCSDS header / packet properly" or something equally verbose points the user in the correct direction and keeps this error distinct from the other "Insufficient packet length" error(s)

ait/core/server/handler.py
```diff
@@ -125,18 +123,24 @@ def handle(self, input_data):
...
-            msg += ' Available packet APIDs are {}'.format(self.packet_types.keys())
             ait.core.log.info(msg)
             return
         packet_name = self.packet_types[packet_apid]
         tlm_dict = tlm.getDefaultDict()
         packet_uid = tlm_dict[packet_name].uid
 
         packet_data_length = int(binascii.hexlify(packet[10:12]), 16) + 1
+        if (len(packet) < 12 + packet_data_length):
+            ait.core.log.info('CCSDSPacketHandler: Insufficient packet length.')
```
MJJoyce: Same as above. "Insufficient packet length" isn't wrong but more details would be better!

ait/core/server/stream.py
```diff
@@ -55,6 +55,7 @@ def process(self, input_data, topic=None):
         """
         Invokes each handler in sequence.
         Publishes final output data.
+        Returns None if a handler cannot process input data
```
MJJoyce: At the risk of being a bit pedantic, `process()` returns None even if it successfully processes input data. I would like to see a bit of elaboration about None causing the handler string to terminate and no data to be published, etc.

MJJoyce:
@aywaldron Would be great to get your review on this as well since you're much more familiar with the code than I am!

ait/core/server/handler.py
aywaldron: Also its confusing that this is the exact same error message as above - that won't be helpful to users trying to identify what the exact problem is.

ait/core/server/stream.py
```diff
@@ -63,7 +64,11 @@ def process(self, input_data, topic=None):
         """
         for handler in self.handlers:
             output = handler.handle(input_data)
-            input_data = output
+
+            if output:
+                input_data = output
+            else:
+                return
```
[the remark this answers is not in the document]
aywaldron: I think there should be something here in addition to the log messages within the handler.

seanlu99:
@aywaldron I fixed all of your change requests. Thanks for the advice.

aywaldron:
Thanks @seanlu99! One last comment - I think it would cleaner to determine the type of the handler dynamically rather than checking for individual types. There are only 2 types for now, but that will change in the future and we shouldn't have to go in and edit the code to check for additional types later.

ait/core/server/handler.py
```diff
@@ -59,8 +59,7 @@ def __init__(self, input_type=None, output_type=None, **kwargs):
             **kwargs:
                 packet:   (required) Name of packet, present in default tlm dict.
         Raises:
-            ValuError:    If packet is not present in kwargs.
-                          If packet is specified but not present in default tlm dict.
```
aywvt: This should be added back in, right?

ait/core/server/handler.py
```diff
@@ -125,7 +125,11 @@ def handle(self, input_data):
         Returns:
             tuple of packet UID and packet data field
         """
+
         packet = bytearray(input_data)
+        if (len(packet) < 13):
+            ait.core.log.info('CCSDSPacketHandler: Received packet length is less than minimum of 7 bytes.')
```
aywaldron: What's the reason for the discrepancy between checking for length of 12 and saying the minimum is 7? Or am I missing something here?

seanlu99: The first 6 bytes are always the same. Possibly extra info coming from ZMQ.

seanlu99:
@aywaldron I believe I addressed all the change requests. Is there anything else you would like me to change?

ait/core/server/handler.py
MJJoyce: Hey @seanlu99, let's dig into this a bit more. I'd like a more concrete answer to what the 6 bytes are, why they're ignored, etc.

Could you post some info on what they are / what you're seeing?

ait/core/server/handler.py
```diff
@@ -136,8 +141,12 @@ def handle(self, input_data):
         tlm_dict = tlm.getDefaultDict()
         packet_uid = tlm_dict[packet_name].uid
 
-        packet_data_length = int(binascii.hexlify(packet[10:12]), 16) + 1
+        packet_data_length = int(binascii.hexlify(packet[4:6]), 16) + 1
+        if (len(packet) < 6 + packet_data_length):
```
aywaldron: @seanlu99 Is there any reason you can't change this to the following, by moving the +1 out of `packet_data_length`? I'm just curious why you split it up this way, as it is a little confusing to read the code with the check for `< 7` above and `< 6 + packet_data_length` here. 
```suggestion
        packet_data_length = int(binascii.hexlify(packet[4:6]), 16) 
        if (len(packet) < 7 + packet_data_length):
```

seanlu99: I did it that way because the CCSDS documentation says that "A Space Packet shall consist of at least 7 octets" and the Packet Data Length is "one fewer than the length of the Packet Data Field." But now that you point it out I get why they subtracted 1 from the field. I'll change it to what you suggested.

seanlu99: @aywaldron Wait actually if I do it the way you suggested then I have to do a `+2` on `udf_length` and `udf_start` still has a `+6` in it.

aywaldron:
@seanlu99 and I discussed the checks on the byte lengths and reworked a couple variables to better reflect the CCSDS standard.

aywaldron:
Merged per @MJJoyce

PR #163, opened 2019-06-10
Issue #162 - Added more robust handling of APID masking in CCSDSHandler

### LINKED ISSUE #162: Add more robust handling of APID masking in CCSDSHandler
Most of the time all 11 bits of the APID are not used. To counter this issue, we should allow users to select a specific mask for the APID in the config. For example,
`XXXXX1011XXX: SomePacket`
would indicate that only bits 6-9 represent the APID for SomePacket.

seanlu99:
Resolves #162

### DIFF
--- a/ait/core/server/handler.py
+++ b/ait/core/server/handler.py
@@ -91,9 +91,11 @@ class CCSDSPacketHandler(Handler):
     """
     This CCSDS handler provides a way to accept multiple packet types on a
     single stream and have them be processed. This handler takes a string of
-    packed binary data (packed by struct) containing CCSDS packet data. It then
-    maps the APID of the packet to a packet name from the config. Using this
-    packet name, it returns the UID and user data field.
+    raw binary data containing CCSDS packet data. It maps the APID of
+    the packet to a packet name from the config, and then uses the packet name
+    to get the UID from the default telemetry dictionary. The user data field
+    is extracted from the raw binary data, and a tuple of the UID and user data
+    field is returned.
     """
     def __init__(self, input_type=None, output_type=None, **kwargs):
         """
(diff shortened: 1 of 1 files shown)


ait/core/server/handler.py
```diff
@@ -125,26 +131,63 @@ def handle(self, input_data):
...
         udf_length = packet_data_length - self.packet_secondary_header_length
         udf_start = primary_header_length + self.packet_secondary_header_length
         user_data_field = packet[udf_start:udf_start + udf_length + 1]
-        return pickle.dumps((packet_uid, user_data_field), 2)
\ No newline at end of file
+
+        ait.core.log.info(packet_apid)
+        ait.core.log.info(user_data_field[0])
+        ait.core.log.info(user_data_field[1])
+        ait.core.log.info(user_data_field[2])
```
aywaldron: Should all these logs be in here?

ait/core/server/handler.py
```diff
@@ -125,26 +131,63 @@ def handle(self, input_data):
...
+        # Extract APID from packet
+        packet_apid = str(bin(int(binascii.hexlify(packet[0:2]), 16) & 0x07FF))[2:].zfill(11)
+
+        # Check if packet_apid matches with an APID in the config
+        config_apid = ''
+        for i in self.packet_types:
+            if self.comp_apid(packet_apid, i):
+                config_apid = i
+                break
+        if not config_apid:
```
aywaldron: Personally I would move the comparison to each of `self.packet_types` to your `comp_apid` method, and have it return either the `config_apid` it matches or None if it doesn't match any, then replace this chunk with 
```
config_apid = self.comp_apid(packet_apid)
```
But that's just personal preference, this also works.

seanlu99: I'll change it to that. I think it looks cleaner too.

MJJoyce:
This needs tests added (I'll open a ticket to track) but I think it's worth getting in as-is. Thanks for the PR @seanlu99!!

PR #170, opened 2019-06-18
Issue #169 - Changed pop to get in server plugin creation

### LINKED ISSUE #169: Server.py is removing data from config when creating plugin
_create_plugin function is removing `name`, `inputs`, and `outputs` from server config. This information may be needed in the future (e.g. getting plugin information for accessing database in AIT GUI).

seanlu99:
Resolves #169

### DIFF
--- a/ait/core/server/server.py
+++ b/ait/core/server/server.py
@@ -8,7 +8,7 @@ import ait.core.server
 from stream import PortInputStream, ZMQStream, PortOutputStream
 from broker import Broker
 from ait.core import log, cfg
-
+import copy
 
 class Server(object):
     """
@@ -251,7 +251,7 @@ class Server(object):
         if config is None:
             raise ValueError('No plugin config to create plugin from.')
 
-        name = config.pop('name', None)
+        name = config.get('name', None)
         if name is None:
             raise(cfg.AitConfigMissing('plugin name'))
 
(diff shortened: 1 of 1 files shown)


ait/core/server/server.py
```diff
@@ -286,7 +291,7 @@ def _create_plugin(self, config):
                                 zmq_args={'zmq_context': self.broker.context,
                                           'zmq_proxy_xsub_url': self.broker.XSUB_URL,
                                           'zmq_proxy_xpub_url': self.broker.XPUB_URL},
-                                **config
+                                **other_args
```
aywaldron: Hi @seanlu99, this works, but I would like to propose a different approach. Instead of continuing to pass in the inputs, outputs, and other arguments separately, what if we change the initialization to just take in the entire `**config` as `kwargs`, like
```suggestion
        instance = plugin_class(zmq_args={'zmq_context': self.broker.context,
                                          'zmq_proxy_xsub_url': self.broker.XSUB_URL,
                                          'zmq_proxy_xpub_url': self.broker.XPUB_URL},
                                **config)
```
This would require editing the plugin parent class's initializer and moving the error checks for no inputs/outputs there.

aywaldron: In addition, if we do this, for consistency I'd like to update the stream and handler classes to work the same way. This will also require updating of the relevant unit tests.

Issue #176, opened 2019-07-11
CLTU integration and conversion to AIT cmds


robschneider16:
Work on implementing integration with CLTU and CLTUF files.
A CLTU is similar to an ait cmd but contains a different format and resolution on certain fields.
A CLTUF is similar to an ait sequence.
Work on converting a CLTU to an ait cmd, and CLTUF to an ait sequence.
Eventually integrate support of using CLTUs in place of ait commands. - More info here

PR #177, opened 2019-07-11
Issue #176 - CLTUF to AIT seq converter

robschneider16:
Added a simple script to parse a CLTUF and extract the relevant
fields to construct commands in the nominal ait cmd format.
Currently, parser does not translate opcode to cmd name string,
Nor does it check the command arguments in the CLTU against whats
defined in cmd.yaml


robschneider16:
Currently, the script does not check the CLTU command against the cmd.yaml, so it does not convert the opcode to a string and does not check # of arguments. Should that be implemented? or should we create a cltu api that can replace the cmd api? or some functions to convert to and from? or have a to_cltu method in the cmd api? Thoughts?

robschneider16:
example on how to run. Not sure if i can provide sample input file on public GH

./ait_cltuf_read.py ./AIT_sequence.cmdpkt --output test.txt

MJJoyce:
@robschneider16, we chatted about this off list, so closing it out for now. Please keep the branch around in case we end up needing it in the future!

PR #179, opened 2019-07-16
Issue #178 - Added commit statement and quotation marks to SQLiteBackend

### LINKED ISSUE #178: SQLiteBackend in ait.core.db has bugs
1. Quotation marks are missing around table names in sql statements for the _create_table and insert functions. This causes sql statements with table names with underscores to fail, such as "1553_HS_Packet".
2. A commit statement is missing at the end of the insert function.
3. If a database name is not specified in DataArchive plugin, AIT fails to automatically create a database and connect to it. The default database name needs to be changed to 'ait.db' instead of 'ait'. Also, connect should automatically call create, not the other way around. Also, connect and create should be able to get the config key "database.dbname" to keep uniformity with InfluxDB.

seanlu99:
Resolves #178

### DIFF
--- a/ait/core/db.py
+++ b/ait/core/db.py
@@ -473,7 +473,7 @@ class SQLiteBackend(GenericBackend):
                 should be made.
         '''
         cols = ('%s %s' % (defn.name, self._getTypename(defn)) for defn in packet_defn.fields)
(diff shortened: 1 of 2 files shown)


aywaldron:
Hey @seanlu99 , if you check the build details it looks like you'll need to update a couple tests as part of this ticket to get the build to pass.

PR #186, opened 2019-07-29
Issue #185 - Fixed CCSDS Packet Handler APID comparison

### LINKED ISSUE #185: CCSDS Packet Handler does not always find correct APID from config
The comp_apid function in the CCSDS packet handler returns the first APID in the dictionary from the config. It should return the first matching APID from the config.

seanlu99:
Resolves #185 
Also removes extraneous byte array conversion in CCSDS packet handler.


MJJoyce:
+1 to merge as is. Let's make sure we get some tickets open to follow up with tests for this sort of stuff

PR #191, opened 2019-08-22
Issue #68 - Add sequence documentation and example

### LINKED ISSUE #68: Add (better) documentation for sequence functionality
The sequence functionality is largely undocumented. We should update the documentation to have a sequence-specific section so users have a clear understanding of what the toolkit supports

MJJoyce:
Resolve #68

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -5,7 +5,7 @@ default:
         port: 3075
 
     sequence:
-        directory: ../../seq/
+        directory: ../sequences
 
     script:
         directory: ../../script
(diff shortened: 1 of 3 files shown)


aywaldron:
Is there an example of how to use a sequence txt file in a script?

MJJoyce:
@aywaldron, I don't think we have an example of that specifically. You can call out via `subprocess` in the scripts and run anything, so technically you could call `ait-seq-send` to send one up. I don't know that I've ever run into that use case but it might not hurt to add docs to the scripts API page elaborating that it's running Python scripts, so anything you want to do with that should be possible.

aywaldron:
@MJJoyce Ah, so the txt file is used by `ait-seq-send` then? That wasn't clear to me so maybe we can add mention of that to these docs.

Issue #167, opened 2019-06-17
Add !include capability to cmd.yaml


aywaldron:
Allow for organization of commands into multiple yaml files by adding the !include capability to the cmd.yaml, as available to other config files.

MJJoyce:
This was added in #200

PR #196, opened 2019-09-03
Issue #167 - Add !include capability to cmd.yaml

Futabay:
* Created additional command yaml file `core_set_op_mode.yaml` - command was copied from [AIT-Core Users Guide](https://ait-core.readthedocs.io/en/latest/command_intro.html)
* Updated `cmd_schema.json` and `cmd.py` to add `!include` capability
* I see an issue when using `cmd.encode()` with the added command. Please let me know if this is something I should investigate.
Screen Shot 2019-09-02 at 11 51 51 PM


MJJoyce:
Hey @Futabay,

I think you're seeing that error since it's trying to pull arguments during encoding and none were passed in on creation of the command.

I think the following would work. Give it a shot and let me know
```
test = cmd.create('CORE_SET_OP_MODE', 'SAFE')
test.encode()
```

Futabay:
Ah, yes it worked! Thank you!
Screen Shot 2019-09-03 at 7 42 00 PM

PR #200, opened 2019-09-08
Issue #167 - Add !include capability to cmd.yaml

Futabay:
* Created additional command yaml file `core_set_op_mode.yaml` - command was copied from [AIT-Core Users Guide](https://ait-core.readthedocs.io/en/latest/command_intro.html)
* Updated `cmd_schema.json` and `cmd.py` to add `!include` capability

FYI, below is the `ait-yaml-validate --cmd`, `Cmd.encode()`, and `Cmd.decode()` outputs:
Screen Shot 2019-09-08 at 4 54 40 AM

### DIFF
--- a/ait/core/cmd.py
+++ b/ait/core/cmd.py
@@ -462,7 +462,9 @@ class CmdDict(dict):
             else:
                 stream        = content
 
-            for cmd in yaml.load(stream):
+            cmds = yaml.load(stream)
+            cmds = handle_includes(cmds)
+            for cmd in cmds:
                 self.add(cmd)
 
             if type(stream) is file:
@@ -496,6 +498,22 @@ def getMaxCmdSize():
     removes 1 word for CCSDS header (-1)
     """
     return (MAX_CMD_WORDS - 1) * 2
+
+
+def handle_includes(defns):
+    '''Recursive handling of includes for any input list of defns.
+    The assumption here is that when an include is handled by the
+    pyyaml reader, it adds them as a list, which is stands apart from the rest
+    of the expected YAML definitions.
+    '''
+    newdefns = []
+    for d in defns:
+        if isinstance(d,list):
+            newdefns.extend(handle_includes(d))
+        else:
+            newdefns.append(d)
+
+    return newdefns
 
 
 def YAMLCtor_ArgDefn(loader, node):
--- a/ait/core/cmd.py
+++ b/ait/core/cmd.py
@@ -509,6 +509,15 @@ def YAMLCtor_CmdDefn(loader, node):
     fields['argdefns'] = fields.pop('arguments', None)
     return createCmdDefn(**fields)
 
+def YAMLCtor_include(loader, node):
+    # Get the path out of the yaml file
+    name = os.path.join(os.path.dirname(loader.name), node.value)
+    data = None
+    with open(name,'r') as f:
+        data = yaml.load(f)
+    return data
+
+yaml.add_constructor('!include' , YAMLCtor_include)
 yaml.add_constructor('!Command' , YAMLCtor_CmdDefn)
 yaml.add_constructor('!Argument', YAMLCtor_ArgDefn)
 yaml.add_constructor('!Fixed'   , YAMLCtor_ArgDefn)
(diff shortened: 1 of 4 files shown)


MJJoyce:
We're going to want to define the included commands with the same format that we do in the telemetry dictionary for simplicity. For the telemetry stuff we would:

Define the `include` as a list element in the dictionary file
```
...

    - !Field
      name:  footer
      type:  MSB_U32
      value: 0x89ABCDEF

- !include ccsds_header.yaml

```

Include each packet definition as a list item in the included file
```
  
- !Packet
  name: CCSDS_HEADER
  fields:
    - !Field
      name:       version
      desc:       Indicates CCSDS Version-1 (does not change)
      bytes:       0
      type:       U8
      mask:       0xE0

....
```

With the current approach here you can only have a single command in the included file. So you'd want something like this.

```
# An example command for setting the operation
# mode of an instrument.
- !Command
  name:      CORE_SET_OP_MODE
  opcode:   0x0006
  subsystem: CORE
  desc:      |
    This command sets the operational mode.
  arguments:
    - !Argument
      name:  mode
      desc:  Mode
      units: none
      type:  U8
      bytes: 0
      enum:
        0: SAFE
        1: IDLE
        2: SCANNING
        3: SCIENCE
```

The different in format handling means the validation isn't catching errors that we would expect. For instance, if I set the opcode on `CORE_SET_OP_MODE` to `BADOPCODE` it still passes validation. 

If I do that with the above command format change note that the error is caught when validating the specific file but the command dictionary load doesn't handle the "correct" format properly.

```ait-yaml-validate --cmd --yaml config/core_set_op_mode.yaml```
Screen Shot 2019-09-09 at 18 05 57

``````ait-yaml-validate --cmd```
Screen Shot 2019-09-09 at 18 07 01

This is because CmdDict doesn't handle these nested lists like TlmDict does. When loading elements TlmDict uses a helper to remove the nested lists
https://github.com/NASA-AMMOS/AIT-Core/blob/master/ait/core/tlm.py#L1056

Could you take a stab at updating this @Futabay?

Futabay:
Sure, I will work on that!

MJJoyce:
Just a note that I opened #201 since it's semi-related to the above. Don't end up fighting with that issue when handling the format changes mentioned above. In other words, it's expected that the validation won't be handling stuff completely properly until #201 is resolved as well due to the way the schema / validation checks work.

MJJoyce:
Hey @Futabay,

Thanks for pushing some updates here but I think there's a few more to go. It looks like the build is breaking for a few reasons that aren't related to the #201 stuff.

The include definitions end up nested when they're imported. If you look at the the `TlmDict.load` function you can see it's using a helper caleld `handle_includes` to fix this.
https://github.com/NASA-AMMOS/AIT-Core/blob/master/ait/core/tlm.py#L1040
https://github.com/NASA-AMMOS/AIT-Core/blob/master/ait/core/tlm.py#L1124

There's some other changes we could make in the validation code that's semi-related to this but I think that's sufficient to fix your broken tests. Let me know if you run into issues!

Futabay:
Hi @MJJoyce 
Thank you so much for your review and comments. I'm sorry that I missed the parts. I will work on them.

Futabay:
Hi @MJJoyce
I added `handle_includes` to `cmd.py`. It seems that the nested list issue was resolved, but still shows error message when I run `ait-yaml-validate --cmd`. Based on your comment, my understanding is that this error message won't be gone until #201 is resolved. Please let me know if I missed something on this issue. FYI, I am attaching my quick test outputs.

Before:
Screen Shot 2019-09-22 at 8 06 38 PM

After:
Screen Shot 2019-09-22 at 8 08 02 PM

btw, sorry for the merge requests...

MJJoyce:
Hey @Futabay,

Thanks for the updates. It looks like this covers the majority of the issues. The test failing that you mention isn't actually covered in #201 but that's largely because I did a poor job of explaining the issue. I've opened a different ticket to cover this (#206) and I've written up a working solution to that based on the changes in this PR.

So I'm +1 to merge even though the tests are breaking here since the changes here aren't to blame.

Can you please rebase out the merge commits so we have a clean history prior to merge?

Thanks again!

PR #207, opened 2019-10-02
Issue #206 - Fix command duplicate opcode / name validation

### LINKED ISSUE #206: CmdDict should fail loudly when encountering duplicate opcode / command name
The current `CmdDict` load doesn't fail loudly when it encounters a duplicate command or opcode during config load. The current approach just overwrites existing references. This also affects the way we do command validation and causes problems when updating the approach to match the telemetry dictionary validation style to support nested command files.

- Update `CmdDict.add` so duplicate opcodes or command names result in a raised exception (similar to current `TlmDict` handling).
- Update command dictionary load during validation to create a `CmdDict` object instead of loading the YAML "manually". This is necessary to support `include`'d commands.
- Update command validation checks to handle raised exceptions for duplicate opcode and command name tests.

MJJoyce:
Update CmdDict.add so duplicate opcodes or command names result in a
raised exception. This mirrors the TlmDict handling.

Update command dicitonary load during validation to create a CmdDict
object instead of loading the YAML "manually." This is necessary to
support included commands

Update command validation checks to handle raised exceptions for
duplicate opcode and command names. Fix error in duplicate opcode test
so commands don't have duplicate names as well.

Resolve #206

### DIFF
--- a/ait/core/cmd.py
+++ b/ait/core/cmd.py
@@ -397,8 +397,19 @@ class CmdDict(dict):
 
     def add(self, defn):
         """Adds the given Command Definition to this Command Dictionary."""
-        self[defn.name]            = defn
-        self.opcodes[defn._opcode] = defn
+        if defn.name not in self:
+            self[defn.name] = defn
+        else:
+            msg = "Duplicate Command name '%s'" % defn.name
+            log.error(msg)
+            raise util.YAMLError(msg)
+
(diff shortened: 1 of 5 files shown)


MJJoyce:
Merging this without +1s so we can fix the broken build / other PRs. If someone finds a problem we'll handle it in another ticket.

PR #209, opened 2019-10-04
Issue #208 - Plugin restructure and cleanup

### LINKED ISSUE #208: Plugin structure cleanup
With the addition of the OpenMCT plugin from #204, it seems like a good time to restructure the way we currently group our default plugins.

### LINKED ISSUE #210: Remove test files from generated API docs
Our test files are currently included in our docs build so we end up with useless API pages such as https://ait-core.readthedocs.io/en/latest/ait.core.test.html

Remove these files and update the docs generation guide and githook to fix this.

MJJoyce:
Restructured our plugins and did a bit of cleanup in the plugin code.

Resolve #208 
Resolve #210


MJJoyce:
Note, still need to test a bit to make sure I didn't break any expected imports, etc.

MJJoyce:
Added #210 documentation cleanup in here (since I needed to mess around with it anyway for other updates) and included a reference to new plugin API pages to server page.

ait/core/server/plugins/openmct.py
```diff
@@ -131,45 +115,27 @@ def _checkConfig(self):
...
 
-            elif topic == "log_stream":
-                self.process_log_msg(input_data)
-                processed = True
+        Supported message topics are "telem_stream"
 
-        if not processed:
-            raise ValueError('Topic of received message not recognized as telem or log stream.')
+        """
+        if "telem_stream" in topic:
```
aywaldron: Do we want to add telem streams to this plugins config like we did with the GUI plugin? (see [here](https://github.com/NASA-AMMOS/AIT-GUI/blob/1798cfb6971d36edca38ccbcc85eccbf6c3768c4/ait/gui/__init__.py#L256))

MJJoyce: ya that's a good idea. I dropped the log message handling from the plugin since I don't see it getting used (or being used in the future) but it definitely makes sense to better handle the telemetry streams.

Issue #194, opened 2019-08-28
Add support for Extensions to seq.py


MJJoyce:
`seq.py`  doesn't allow users to expand core classes like some of the other core modules. We should update this so users can make updates to sequence encoding formats if necessary.

MJJoyce:
This was added ages ago but the Resolve text was typoed ...

PR #230, opened 2019-12-12
Issue #194 - Add support for Extensions to seq.py

bruceplai:
Added support for Extensions to seq.py


MJJoyce:
This doesn't work as is so I'm closing it out. I'll leave the branch up for when this is revisited.

Issue #248, opened 2020-02-26
'ait-seq-send' Test


Futabay:
0. Precondition
    1. AIT is installed with virtual environment
    2. Virtual environment is started by typing workon ait on your working terminals
    3. .rts file needs to be available. The file was created based on https://github.com/NASA-AMMOS/AIT-Core/issues/168#issuecomment-505651260

1. Go to  a terminal, invoke `ait-seq-send` command 
```
(ait) $  ait-seq-send </path/to>/test.rts
2020-02-24T21:36:07.524 | WARNING  | Config parameter command.history.filename specifies nonexistent path </some/path/to>/cmdhist.pcap
2020-02-24T21:36:07.527 | WARNING  | Config parameter dsn.cfdp.mib.path specifies nonexistent path </some/path/to>/AIT-Core/config/mib
2020-02-24T21:36:07.529 | WARNING  | Config parameter dsn.cfdp.datasink.outgoing.path specifies nonexistent path </some/path/to>/AIT-Core/ait/dsn/cfdp/datasink/outgoing
2020-02-24T21:36:07.531 | WARNING  | Config parameter dsn.cfdp.datasink.incoming.path specifies nonexistent path </some/path/to>/AIT-Core/ait/dsn/cfdp/datasink/incoming
2020-02-24T21:36:07.532 | WARNING  | Config parameter dsn.cfdp.datasink.tempfiles.path specifies nonexistent path </some/path/to>/AIT-Core/ait/dsn/cfdp/datasink/tempfiles
2020-02-24T21:36:07.534 | WARNING  | Config parameter dsn.cfdp.datasink.pdusink.path specifies nonexistent path </some/path/to>/AIT-Core/ait/dsn/cfdp/datasink/pdusink
2020-02-24T21:36:07.536 | WARNING  | Config parameter data.1553.path specifies nonexistent path /gds/dev/data/LMC-062858/2020/2020-056/downlink/1553
2020-02-24T21:36:07.538 | WARNING  | Config parameter data.bad.path specifies nonexistent path /gds/dev/data/LMC-062858/2020/2020-056/downlink/bad
2020-02-24T21:36:07.540 | WARNING  | Config parameter data.lehx.path specifies nonexistent path /gds/dev/data/LMC-062858/2020/2020-056/downlink/lehx
[...]

Issue #168, opened 2019-06-17
ait_cmd_send should support numeric values for string arguments


nttoole:
With the resolution to #66, the ait_cmd_send script will always report an error when a numeric value is provided for a string argument, even if the argument is surrounded by quotes.

Sample command with one string argument:

```
- !Command
 name:      SEND_STR_ARG
 opcode:    0x0005
 subsystem: CMD
 title: Send string argument value
 desc:      |
   This command tests sending a single string argument.

 arguments:
   - !Argument
     name: str_arg
     desc: String Argument
     units: none
     type: S16
     bytes: [0,15]
```

Sending the command:

```
>> $AIT_ROOT/ait/core/bin/ait_cmd_send.py SEND_STR_ARG 1
2019-06-17T09:51:32.784 | PROGRAM  | /Users/nttoole/dev/ait/git_repos/ait_052919/AIT-Core/ait/core/bin/ait_cmd_send.py SEND_STR_ARG 1
2019-06-17T09:51:32.787 | ERROR    | str_arg value ‘1’ is not a string type.
2019-06-17T09:51:32.787 | PROGRAM  | done.
```
and
```
>> $AIT_ROOT/ait/core/bin/ait_cmd_send.py SEND_STR_ARG "1"
2019-06-17T09:51:32.784 | PROGRAM  | /Users/nttoole/dev/ait/git_repos/ait_052919/AIT-Core/ait/core/bin/ait_cmd_send.py SEND_STR_ARG 1
2019-06-17T09:51:32.787 | ERROR    | str_arg value ‘1’ is not a string type.
2019-06-17T09:51:32.787 | PROGRAM  | done.
```

The issue is related to how the command arguments are parsed, where numeric values are parsed as Python numeric primitives with no knowledge of the command and its argument definition.  
When attempting to validate the argument with command's argument definition, there is a type conflict.

nttoole:
After the fix: 

Sending integer for string argument:
```
MT-202661{nttoole}% $AIT_ROOT/ait/core/bin/ait_cmd_send.py SEND_STR_ARG 1
2019-06-25T10:58:39.517 | PROGRAM  | /Users/nttoole/dev/ait/git_repos/ait_052919/AIT-Core/ait/core/bin/ait_cmd_send.py SEND_STR_ARG 1
2019-06-25T10:58:39.520 | COMMAND  | Sending to 127.0.0.1:3075: SEND_STR_ARG 1
[...]

PR #254, opened 2020-03-31
kkallas python3 test update

Futabay:
ait/core/dtype.py need further investigation (Issue #248). This issue seems to be related to #168 (comment)


nttoole:
@Futabay, the ait/core/dtype.py fix is included in my set.  ;)

Issue #264, opened 2020-04-09
Reverse indexing cleanup in ait_seq_print and ait_seq_decode


MJJoyce:
Minor cleanup of the following would be good

```
        # Parse the filename for the applicable information
        parts = os.path.basename(filename).split('_')
        l = len(parts)
        desc = os.path.splitext(parts[l-1])[0]
        seqid = parts[l-2]
        subsys = parts[l-3]
```

MJJoyce:
Resolve in #271

PR #271, opened 2020-04-17
Issue #264 - Reverse indexing cleanup

Futabay:
Reverse indexing cleanup in ait_seq_print, ait_seq_decode, and ait-seq-encode.


MJJoyce:
+1, looks good!

PR #274, opened 2020-04-22
Issue #143 - CmdApi utilizes newly added server command entry point

### LINKED ISSUE #143: Update CmdAPI to default to a default server output stream / ZMQ instead of UDP
The CmdAPI should output commands to the server output streams so that it can handle output as defined in the config. This will make the server the only input/output configuration point. This will require defining a "default" output stream option in the config / code so that commands are dumped where the user expects (or a sane default if nothing is specified).

nttoole:
- introduced concept of a command topic, which the server automatically create
- server will connect this topic to at most one outbound stream
- outbound stream config updated to include new field designating it
- if not outbound stream declares itself, then server will self-assign the first out stream
- updated server architecture docs to document the above
- CmdApi updated to send command+args to either ZMQ topic (default) or UDP.
- updated ait_cmd_send and ait_seq_send to include new options for specifying topic or udp.

Resolve #143

### DIFF
--- a/ait/core/server/broker.py
+++ b/ait/core/server/broker.py
@@ -25,6 +25,10 @@ class Broker(gevent.Greenlet):
         self.XPUB_URL = ait.config.get('server.xpub',
                                         ait.SERVER_DEFAULT_XPUB_URL)
 
+        ## Name of the topic associated with external commands
+        self.command_topic = ait.config.get('command.topic',
+                                        ait.DEFAULT_CMD_TOPIC)
+
         gevent.Greenlet.__init__(self)
 
     def _run(self):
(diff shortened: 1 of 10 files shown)


ait/core/server/broker.py
```diff
@@ -81,6 +87,52 @@ def _subscribe_all(self):
...
+    def _subscribe_cmdr(self):
+        """
+        Setup for the appropriate outbound stream that is configured to
+        accept command messages.  If no stream is specified, it looks
+        for the first outbound stream.
+        """
+
+        ## Only continue if command topic was supplied
+        if not self.command_topic:
+            return
```
MJJoyce, line 102: Let's dump an error message here. This seems like a bad state to end up in quietly

ait/core/server/broker.py
```diff
@@ -25,6 +25,12 @@ def __init__(self):
         self.XPUB_URL = ait.config.get('server.xpub',
                                         ait.SERVER_DEFAULT_XPUB_URL)
 
+        ## Name of the topic associated with external commands
+        self.command_topic = ait.config.get('command.topic',
+                                        ait.DEFAULT_CMD_TOPIC)
+        ## Controls if server selects a default if one not speicified
+        self.command_stream_select_defaut = True
```
MJJoyce, line 32: This isn't configurable so I vote we drop this. Worst case we can add it similar functionality in the future if this is a use case we want to cover (currently I don't think it is)

ait/core/server/broker.py
```diff
@@ -81,6 +87,52 @@ def _subscribe_all(self):
...
+        ## Warn about multiple matches
+        if cmd_stream and len(cmd_streams) > 1:
+            log.warn('Multiple output streams found with {} field enabled, '
+                     '{} was selected as the default.'.format(
+                     cmd_sub_flag_field, cmd_stream.name))
+
+
+        ## Do we want to make a guess for the command stream?
+        ## Just grab the first output stream
+        if not cmd_stream and self.command_stream_select_defaut:
```
MJJoyce, line 123: Per the above, I say we drop the `and self.command_stream_select_default` from the check

ait/core/server/broker.py
MJJoyce: On second thought, do we even care about this? It's not really possible that we end up triggering this block. If something really odd happened and `ait.DEFAULT_CMD_TOPIC` wasn't available I think the failure when we try to subscribe is acceptable. Thoughts?

config/config.yaml
```diff
@@ -3,6 +3,8 @@ default:
         history:
             filename: ../../cmdhist.pcap
         port: 3075
+        topic: __commands__
+        zmq_conn_wait: 1
```
MJJoyce, line 7: I think we can drop these from the default config and stick with documenting them only. We pull defaults for them and I really can't imagine a situation where 99% of users want to change them. The slimmer the default config the better.

ait/core/server/broker.py
nttoole: I prefer to keep _some_ level of sanity checking in place - even when the design says they shouldn't be necessary.  So I went ahead and added the error log message.

MJJoyce: Works for me!

MJJoyce:
Looks good to me!

PR #282, opened 2020-04-28
Issue #281 - Update CmdAPI with None udp_dest default

### LINKED ISSUE #281: Default CmdAPI dest attribute to None
I missed this when reviewing #274. We want the default CmdAPI behavior to be using the the 0MQ port. Let's updated `udp_dest` to be `None` instead of requiring specifying `None` for this parameter when you want to use the 0MQ port. This makes it clearer when someone is deviating from the 0MQ default by requiring `udp_dest` to be called out as well.

Current implementation and call structure:
```
def __init__ (self, udp_dest, cmddict=None, verbose=False, cmdtopic=None):
    
    ...
```

Example of the current use in `ait-cmd-send`
```
    ## If UDP enabled, collect host/port info
    if udp:
        if host is not None:
            dest = (host, port)
        else:
            dest = port

    cmdApi  = api.CmdAPI(dest, verbose=verbose, cmdtopic=topic)
```

Updated version:
```
def __init__ (self, udp_dest=None, cmddict=None, verbose=False, cmdtopic=None):
    
    ...
```

MJJoyce:
Update CmdAPI's upd_dest argument to default to None. Instantiation of
CmdAPI without additional arguments will result in a connection to the
C&DH global 0MQ command port.

Update CmdAPI instantiations in ait-[cmd|seq]-send scripts to take the
keyword argument into consideration.

Resolve #281

### DIFF
--- a/ait/core/api.py
+++ b/ait/core/api.py
@@ -101,7 +101,7 @@ class CmdAPI:
     Datagram Protocol (UDP) packets if udp_dest argument is set,
     else, command is sent via ZMQ topic.
     """
-    def __init__ (self, udp_dest, cmddict=None, verbose=False, cmdtopic=None):
+    def __init__ (self, udp_dest=None, cmddict=None, verbose=False, cmdtopic=None):
 
         if cmddict is None:
             cmddict = cmd.getDefaultCmdDict()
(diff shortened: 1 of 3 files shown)


MJJoyce:
@nttoole, your thoughts on this since you were just making related changes would be awesome.

Issue #285, opened 2020-06-29
Add logging to UTC Leapsecond handling


MJJoyce:
The leapsecond load / processing is a bit opaque at the moment. Adding a few logs in the process would help clear up what's going on when AIT boots up and processes the data file or downloads updates.

PR #288, opened 2020-07-23
Issue #285 - Added more log info into dmc.py

acliu88:
Updated dmc.py with more logging information when pulling, loading, or generating leapsecond.dat file.


ait/core/dmc.py
```diff
@@ -302,6 +304,10 @@ def _load_leap_second_data(self):
                     log.warn('Continuing with out of date leap second data')
                 else:
                     raise ValueError('Could not load leap second data')
+        else:
+            t = self._data['valid']
+            t.strftime('%m/%d/%Y')
+            log.info('Leapseconds data valid until %s', t)
```
MJJoyce, line 310: This is going to print the default format for `t`. Something like this:
```
>>> print(ait.core.dmc._data['valid'])
2020-12-28 00:00:00
```

You need to store the return value from `strftime` to change the format.

PR #311, opened 2020-12-11
Database API Result Handling and SQLite bug fixes and improvements

### LINKED ISSUE #303: Database API query and result abstraction
The database APIs provide a baseline abstraction for interfacing with a configured data store but functionality beyond that is lacking. It would be good to see abstractions extended for queries (especially over a time range) and return results so that users and internal functionality and interface with the datastore consistently without concern for specific DB idiosyncrasies (i.e., the entire point of the API ...). Changes here would greatly improve OpenMCT integration and the GUI playback functionality which has been perpetually back-burnered waiting for this.

### LINKED ISSUE #307: SQLite Backend Simplification
The SQLite backend implementation is our baseline / example database implementation. Given the on going work to change the database result set handling, amongst other things, it seems reasonable to review the existing implementation and work through other changes. 

The existing type handling isn't great but seems to largely work in its current state. I'm not sure that the original implementation was ever tested heavily since it was really more "proof of concept".  The result set generalization is likely going to change this code. Since we're going to make changes here, I'd suggest that instead we just store packets as BLOBs in our SQLite abstraction. It certainly reduces potential functionality of the SQLite backend but it simplifies much of the reconstruction and schema creation code (again, which wasn't really heavily tested to my knowledge), and is more than adequate for an example / proof of concept interface.

### LINKED ISSUE #310: DB API Sqlite connection and creation handling fails to properly cover cases
The current SQLite handling of `connect` and `create` don't really cleanly separate the connect/create responsibilities. They also fail to properly create a database when a custom datastore name is defined in config. Both of these need cleaned up.

MJJoyce:
This is a bit of a sweeping rework to parts of the database APIs to
provide a unification to result formatting. During testing a few issues
were run into specific to the SQLiteBackend implementation that resulted
in a complete rework to the data storage format. These changes are
related enough that I'm grouping them under a single pull request.
Related issues:

Issue #303 - Database API query and results abstraction
Issue #307 - SQLite Backend Simplification
Issue #310 - DB API Sqlite connection and creation handling fails
    to properly cover cases

Add AITDBResult class to provide a minimal wrapper around all database
query results. Any API that executes a query and returns results now
returns an AITDBResult object. Users can access the queries execute, any
errors encountered, and the query results (either raw or as processed
Packets depending on the interface) through this object.

Add a `query_packets` function to the backends. These functions take an
optional list of packet types and start / end times and return
time-ordered Packet objects for all results. This is likely the primary
interface that users and AIT will use to access data in a
backend-agnostic way.

Add a `create_packet_from_result` classmethod to the backends. The
backends use these to parse a given query result item into an AIT
packet. Previously only the Influx backend had a similar function (it is
now marked deprecated). These also provide users a helpful interface
for converting results to packets as part of their custom queries.

Completely rework the SQLite backend storage format to use BLOB to store
packets instead of splitting fields out into individual columns.
Ticket #307 provides an explanation for why this was done. In short,
the SQLite support is really meant for demonstration purposes and to
provide a baseline implementation that other backends can reference.
Users are heavily encouraged to NOT use SQLite as their primary mission
backend. Now seemed like a good time to make this switch given the impact
on the existing code from changes for #303 and the lack of thorough testing
for the previous format and type handling.

Fix the poor SQLiteBackend handling of connect / create and fix the
issue where custom named databases weren't being properly configured.
SQLiteBackend.connect now "connects" to the datastore and checks if the
database previously existed. If not, then the table structure needs
handled. SQLiteBackend.create now only creates tables instead of the
previously incorrect hybrid of functionality.

Resolve #303 
Resolve #307 
Resolve #310

### DIFF
--- a/ait/core/db.py
+++ b/ait/core/db.py
@@ -21,10 +21,14 @@ commands and telemetry with several backends.
 from abc import ABCMeta, abstractmethod
 import datetime as dt
 import importlib
+import itertools
 import math
+import os.path
+
+import sqlite3
 
 import ait
-from ait.core import cfg, cmd, dmc, dtype, evr, tlm
+from ait.core import cfg, cmd, dmc, dtype, evr, log, tlm
 
 
 # Backend must implement DB-API 2.0 [PEP 249]
@@ -48,7 +52,7 @@ def create(database, tlmdict=None):
     """
     if tlmdict is None:
         tlmdict = tlm.getDefaultDict()
-    
+
     dbconn = connect(database)
 
     for name, defn in tlmdict.items():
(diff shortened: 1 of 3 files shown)


MJJoyce:
@nttoole, There's a good bit of changes here so another pair of eyes would be much appreciated.

In terms of relevance to the other changes you're making:
- Each backend has a `query_packets` method for querying one or more packet types over a time range.
- Results are returned in the `AITDBResult` wrapper and can be accessed via `object.get_packets()`. They're time ordered.
- If you need access to the time value for each packet you can pass `yield_packet_time=True` to the `query_packets()` call so the generator returns the time value as a `datetime.datetime` object as well.

Rough example call:
```
import datetime as dt
import ait.core.db as db

be = db.SQLiteBackend()
be.connect()

start = dt.datetime(2020, 12, 2, 0, 40, 8)
end = dt.datetime(2020, 12, 2, 0, 50, 9)

res = be.query_packets(start_time=start, end_time=end, yield_packet_time=True)

for t, pkt in res.get_packets():
    # Do cool stuff
```

MJJoyce:
Oops, tests are breaking because I didn't mock out all the imports for other libraries. I'll update. These run clean locally where I have the required dependencies installed for whatever that is worth =)

MJJoyce:
Updated the tests to mock out the backend driver import where possible. Skipped two tests if we can't import the driver since  trying to mock the nested import of an Exception class wasn't working. There might be a simple way to handle that but I couldn't quite figure it out so I skipped instead.

MJJoyce:
@nttoole,

PR updated to address your comments. I've left the changes in an unsquashed commit for now to simplify review.

ait/core/db.py
```diff
@@ -348,13 +463,140 @@ def query(self, query, **kwargs):
...
+
+                for t, pkt in pkt_conv:
+                    if kwargs.get('yield_packet_time', False):
+                        yield (t, pkt)
+                    else:
+                        yield pkt
+
+        return AITDBResult(
+            query=query_string,
+            packets=influx_results_gen(db_res, **kwargs)
```
nttoole, line 587: @MJJoyce Just wondering if there would be any value in having the return object contain the raw results along with the Packet generator?  I don't have a strong opinion either way.

MJJoyce: Hmm I don't think so here. The `query_packets` API is meant to give you back `tlm.Packet`s. If users just want raw query results they can do that through the `query` endpoint. Since running through the raw results would require you to be familiar with the DB result format I don't think it's too much of a stretch to require they also be familiar with the query syntax.

Thoughts?

nttoole (dismissed):

MJJoyce:
Thanks for reviewing @nttoole. I'll squash the fixup commit out and merge. This should cover the pending changes necessary for the OpenMCT plugin changes. Let me know if there's issues when you look at that.

PR #312, opened 2020-12-15
Issue #306 - Support lists in config directory expansion

### LINKED ISSUE #306: Cfg directory expansion fails to handle new server additions
Noticed when working with some of the recent changes to the GUI plugin that when passing a relative directory for the static files no updates were being made to the path.

E.g.
```
    server:
        plugins:
            - plugin:
                name: ait.gui.AITGUIPlugin
                inputs:
                    - log_stream
                    - telem_stream
                outputs:
                    - command_stream
                html:
                    directory: ../gui/
```

The above results in the `html.directory` configuration value not having its path expanded like expected. After reviewing the directory expansion code it's a result of the recursion ignoring any type that isn't a dictionary when attempting to dig down into the structure. We'll need to modify this so the full config structure is parsed.

MJJoyce:
Update the config directory expansion handling to support lists within
dictionary elements. This server rework added elements consisting of
lists of dictionaries, some of which contain elements that could require
path expansion. For example, the GUI plugin looks for static file
directory configuration which likely requires path expansion. These
changes fix the issue in parsing elements of this sort.

Resolve #306

### DIFF
--- a/ait/core/cfg.py
+++ b/ait/core/cfg.py
@@ -63,9 +63,15 @@ def expandConfigPaths (config, prefix=None, datetime=None, pathvars=None, parame
 
             config[name] = cleaned[0] if len(cleaned) == 1 else cleaned
 
-        elif type(value) is dict:
+        elif isinstance(value, dict):
             param_key = name if parameter_key == '' else parameter_key + '.' + name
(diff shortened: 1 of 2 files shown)


MJJoyce:
I got a +1 to merge this from @nttoole when we were chatting yesterday. Onward!

PR #314, opened 2020-12-21
Issue #313 - Update API telemetry handling functionality

### LINKED ISSUE #313: API telemetry handling doesn't use the C&DH server
Necessary changes in the API telemetry handling were overlooked during updates to the C&DH server. Currently the `api.Instrument` interface boots up a webserver to monitor and report instrument telemetry to ground scripts. Given the C&DH server rearchitecting, this should be instead hooking up to the C&DH server and pulling telemetry from there.

--

Brainstorming on changes here:

A similar (but inverse) approach that we took with the commanding port seems like it would work here. We could hook up valid telemetry streams (set either via config or a valid default) as an input stream to this "API stream" and then Instrument / other API functionality can connect to that known socket for information. This would also remove the obnoxious configuration on Instrument and issues when ports clash in the current implementation.

This would likely require modification to the Instrument `__init__` implementation as well as a new API component to replace the webserver.

MJJoyce:
Update API telemetry handling in the Instrument interface. Telemetry
monitoring now handles all valid packets in the telemetry dictionary.
Telemetry data is pulled from the C&DH server via the default telemetry
topic.

Server code has been updated to create a new input stream that takes
input from other valid telemetry streams and emits telemetry data under
the "default telemetry topic" so other code can listen for these data.
Input streams are either set via config or are defaulted, if possible, to
all input streams that emit data in the necessary format (PacketHandler
and CCSDSPacketHandler being the current ones).

Tests have been updated. Global AIT config has been updated so the
"default telemetry topic" can be pulled via ait.DEFAULT_TLM_TOPIC.

Resolve #313

### DIFF
--- a/ait/core/api.py
+++ b/ait/core/api.py
@@ -40,6 +40,7 @@ import collections
 import inspect
 import json
 import os
+import pickle
 import socket
 import time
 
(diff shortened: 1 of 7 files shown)


MJJoyce:
Note this isn't quite ready for prime time yet but I want to get some eyes on it. Tests still need updated / created where possible for the `api.py` changes and docs need updated throughout.

ait/core/api.py
```diff
@@ -554,36 +555,95 @@ def start (self):
...
-                raise TypeError(msg)
+        self._sub = self._cntxt.socket(zmq.SUB)
+        self._sub.connect(sub_url)
+        self._sub.setsockopt_string(zmq.SUBSCRIBE, ait.DEFAULT_TLM_TOPIC)
+
+    def _run(self):
+        try:
+            while True:
+                gevent.sleep(0)
+                raise Exception()
```
nttoole, line 575: @MJJoyce Is this for active dev-debugging?

MJJoyce: Bleh ya, looks like I failed to save properly before pushing the branch up. That's what I get for trying to get up before I ran out the door. I'll fix it up tomorrow along with some other changes to tests and whatnot. Good catch!

ait/core/server/server.py
```diff
@@ -89,6 +95,73 @@ def _load_streams(self):
         if not self.outbound_streams:
             log.warn(err_msgs['outbound'])
 
+    def _create_api_telem_stream(self):
+        """"""
+        stream_map = {'__valid_api_streams': []}
+        COMPATIBLE_HANDLERS = ['PacketHandler', 'CCSDSPacketHandler']
```
nttoole, line 101: @MJJoyce Is this a TODO-make more configurable? Or is this expected to the the standard?

MJJoyce: I think I took out the TODO that I had commented in here. Guess I should have left it in =D

So this list is only the list of handlers that we're using when we try to set a default if nothing is configured. Eventually I'd like to see it (potentially) moved somewhere else that makes more sense or figure out a better way for determining output format compatibility in handlers. Or something else entirely that we can think of that makes more sense. Being able to check handler output compatibility would be nice so we could verify that the configuration is good even when the user is specifying it.

I'm not really what the best approach is though for the time being. I guess we could pull it from constant somewhere so it's easier for users to modify? Of course, if you're going through that much trouble you'd be better off just specifying what streams to use in the config.

MJJoyce:
@nttoole,

Updated the telemetry config attribute name as we discussed offline. If anything else catches your eye as needed updated in here let me know.

ait/core/api.py
```diff
@@ -554,36 +555,94 @@ def start (self):
...
-        self._cmd     = CmdAPI(cmdport)
+        self._pkt_buffs = PacketBuffers()
+        for _, pkt_defn in defns.items():
+            self._pkt_buffs.create(pkt_defn.name)
 
-        self._packets.create(defn.name)
-        pktbuf        = self._packets[defn.name]
-        self._tlm     = UdpTelemetryServer(tlmport, pktbuf, defn)
-        self._tlm.start()
+        self._cmd = CmdAPI(cmdport)
```
nttoole, line 634: @MJJoyce Does this usage still work after we refactored CmdAPI to be stream-aware?

MJJoyce: Ya, looks fine to me. `CmdAPI`'s `__init__`

```
def __init__ (self, udp_dest=None, cmddict=None, verbose=False, cmdtopic=None):
```

`cmdport` defaults to `None` and that gets passed through to `udp_dest` by default.

nttoole: 👍

doc/source/server_architecture.rst
```diff
@@ -130,6 +133,9 @@ Here is an example of how the **server** portion of **config.yaml** should look:
                 outputs:
                     - command_testbed_stream
 
+        api-streams:
```
nttoole, line 136: @MJJoyce Should this be 'api-telemetry-streams:' ?

nttoole (approved):
There's a minor tweak needed in doc/source/server_architecture.rst  ( api-streams -> api-telemetry-streams ) but otherwise looks good

doc/source/server_architecture.rst
MJJoyce: oooof course I missed one ...

PR #322, opened 2021-02-25
Issue #299 - Update server 0MQ data passing

### LINKED ISSUE #299: Update server data passing format
The C&DH server is using `pyzmq`'s `send_string` and `recv_string` when passing data. This is an artifact of the initial implementation for this being done in Python 2.7.x and then converted over. Some of the updates to handle the format changes from the 3.x cutover are less than ideal, including some `eval` calls. 

Ultimately, most handlers and plugins are expecting to receive a bytes-like object as a default. Some will have expected addition processing so they might expect a more complicated message format (e.g., many of the packet handling plugins expect to receive a tuple of the form `(packet id, packet data blob)`). We should update the way we are passing data so the default expectation is met.

The current suggested changes are to update all 0MQ pub-sub calls to use either `send / recv` or `send_pyobj / recv_pyobj`. The payload will be a pickle-serialized tuple of the form `(message topic, bytes-like data object)`. This should keep topic handling simple and flexible like the current form, avoid unnecessary str <-> bytes conversions, and ensure that the previous data form expectations for handlers and plugins is being met (again). This will require updates in Core and GUI.

MJJoyce:
Update server handling of data passing to use 0MQ's
`[send|recv]_multipart` calls instead of `[send|recv]_string`.

Utilities have been added under ait.core.server.utils for encoding /
decoding messages into the proper format to simplify code throughout.

Various hacks for handling the string-ified data have been removed.

Resolve #299

### DIFF
--- a/ait/core/server/plugins/data_archive.py
+++ b/ait/core/server/plugins/data_archive.py
@@ -75,7 +75,7 @@ class DataArchive(Plugin):
             **kwargs:    any args required for connected to the backend
         """
         try:
-            load = pickle.loads(eval(input_data))
+            load = pickle.loads(input_data)
             uid, pkt = int(load[0]), load[1]
             defn = self.packet_dict[uid]
             decoded = tlm.Packet(defn, data=bytearray(pkt))
--- a/ait/core/api.py
+++ b/ait/core/api.py
@@ -47,6 +47,7 @@ import time
 import ait
 import ait.core
 from ait.core import cmd, gds, log, pcap, tlm, util
+import ait.core.server.utils as serv_utils
 
 class APIError (Exception):
     """All AIT API exceptions are derived from this class"""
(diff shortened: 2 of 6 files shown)


MJJoyce:
Note, this is going to need reviewed and messed with a bit more prior to merging

MJJoyce:
@nttoole, I think the 0MQ handling in the OpenMCT plugin looked good but I'll defer to you

ait/core/server/plugins/data_archive.py
```diff
@@ -75,7 +75,7 @@ def process(self, input_data, topic=None, **kwargs):
             **kwargs:    any args required for connected to the backend
         """
         try:
-            load = pickle.loads(eval(input_data))
+            load = pickle.loads(input_data)
```
nttoole, line 78: @MJJoyce I haven't had a deep look, but might input_data already have been de-pickled by the server_util.decode()?  Or are the contents of the message pickled twice before being pushed onto 0Mq?

MJJoyce: I realized I never copy and pasted the answer to this here from our offline chat:

```
currently our packet handlers are pickling the contents before they send them off
so we have (packet id, pickled content) which gets sent to the “client sender”, which pickles it again
```

I'll make sure this gets handled in a follow up ticket

PR #323, opened 2021-03-11
Issue #318 - Fix validation error message sorting issue

### LINKED ISSUE #318: YAML validation failures aren't output properly
It looks like there's an issue with the way the validation scripts are handling error display

This is a snippet of a failed dump
```
   for error in sorted(v.iter_errors(data)):
TypeError: '<' not supported between instances of 'ValidationError' and 'ValidationError'
```

I'm guessing this showed up as part of the Python 3.x transition and got missed. Removing a sort on the error messages resolves the issue:
https://github.com/NASA-AMMOS/AIT-Core/blob/master/ait/core/val.py#L391
```
                for error in sorted(v.iter_errors(data)):
                    msg = "Schema-based validation failed for YAML file '" + self._ymlfile + "'"
```

It's possible that impacts the output formatting but it didn't look off during my testing ...

MJJoyce:
Remove sorting of validation error message to resolve TypeError. Note, we
could implementation our own exception here and address this issue but I
haven't noticed a problem with not sorting the messages.

Resolve #318

### DIFF
--- a/ait/core/val.py
+++ b/ait/core/val.py
@@ -388,7 +388,7 @@ class Validator(object):
 
                 # Loop through the errors (if any) and set valid = False if any are found
                 # Display the error message
-                for error in sorted(v.iter_errors(data)):
+                for error in v.iter_errors(data):
                     msg = "Schema-based validation failed for YAML file '" + self._ymlfile + "'"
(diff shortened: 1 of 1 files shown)


MJJoyce:
I guess I pushed this branch and never made a pull request out of it. Ooppss

PR #349, opened 2021-06-15
Add historical query support to the OpenMCT plugin

### LINKED ISSUE #302: Update OpenMCT plugin to support historical queries
The initial OpenMCT plugin provided support only for realtime telemetry over websockets.
MAIA has an near-immediate need to use OpenMCT plugin, but would need historical queries to be supported by the AIT plugin.
As such, we should add a base implementation of historical queries, which does not need to be too generic at this point, but can be focused on supporting MAIA's needs (i.e. they use the InfluxDb)

nttoole:
The original version of the AIT OpenMct plugin supported realtime telemetry only.
This PR adds historial queries - which is supported by telemetry stored in a database.
The layer uses the newer AIT Database abstraction, and so should work with any supported Database implementation. 

Fixes #302

### DIFF
--- a/ait/core/db.py
+++ b/ait/core/db.py
@@ -549,7 +549,9 @@ class InfluxDBBackend(GenericBackend):
         else:
             etime = dt.datetime.utcnow().strftime(dmc.RFC3339_Format)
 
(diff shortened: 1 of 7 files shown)


MJJoyce (approved):
Looks good!

Issue #354, opened 2021-08-04
Rework table support


MJJoyce:
Our current support for onboard table handling is, frankly, a mess. The encoding is hard coded to only handle a specific header format. Whatever you specify in `table.yaml` is ignored and some of the values are randomly hard coded (e.g., [MagicNumber](https://github.com/NASA-AMMOS/AIT-Core/blob/master/ait/core/table.py#L187)). Many of the abstractions feel a bit all over the place and various parts of the toolkit (e.g., `dtype.py`) aren't sufficiently leveraged.  There's also a bunch of fields (bytes, format, ...) specified in the telemetry dictionary that aren't supported or used in the code at all.

The "decoding" support
https://github.com/NASA-AMMOS/AIT-Core/blob/master/ait/core/table.py#L205

The "encoding" code:
https://github.com/NASA-AMMOS/AIT-Core/blob/master/ait/core/table.py#L353

Also the CLI for encoding / decoding are a mess of mandatory arguments that don't really need to be mandatory arguments.

```
usage: ait-table-encode --fswtabdict config/table.yaml --tabletype targets --tabfile /Users/ays/Documents/workspace/ait-workspace/output/targets_table.txt
```

----

A few things that we should shoot for with a "reimplementation":
- The Core implementation should demonstrate baseline functionality without any (or many) special surprises (like a secretly hard coded header format ...)
- The code should be clean and the abstractions well defined so users can more easily extend the functionality where necessary.
- We should leverage other components in Core whenever possible so functionality is consistent.

MJJoyce:
Didn't close this out in the PR evidently

PR #355, opened 2021-08-04
Issue #354 - Rework table support

MJJoyce:
Rework Core table support to provide users with more consistent and
understandable feature set. Table encoding / decoding functionality has
been (hopefully) simplified via simpler, more flexible APIs, extraction
of hardcoded functionality, and leveraging of other Core functionality
such as dtype.py where possible.

Table encoding and decoding APIs have been cleaned up and made more
flexible so users can pass inputs in expected formats. Some
functionality has been extracted out of table.py into the encode /
decode CLI scripts (where it arguably belonged to begin with) in an
effort to keep Core functionality and abstractions cleaner. For example,
writing a decoded table to a file is no longer a requirement as part of
table decoding.

Column field handling now uses date type information for encoding /
decoding. This maintains consistency with the rest of the toolkit and
will allow easier extension of toolkit functionality when Extension
support is added to table.py.

Encode / decode CLI scripts have been rewritten. Their argument
structures have been simplified in the process.

The default table configuration formation and table.yaml file have been
simplified. A number of attributes have been removed from the dictionary
structures since they weren't supported by the toolkit anyway.

### DIFF
--- a/ait/core/bin/ait_table_decode.py
+++ b/ait/core/bin/ait_table_decode.py
@@ -1,9 +1,9 @@
 #!/usr/bin/env python
-
+#
 # Advanced Multi-Mission Operations System (AMMOS) Instrument Toolkit (AIT)
 # Bespoke Link to Instruments and Small Satellites (BLISS)
 #
-# Copyright 2016, by the California Institute of Technology. ALL RIGHTS
+# Copyright 2021, by the California Institute of Technology. ALL RIGHTS
 # RESERVED. United States Government Sponsorship acknowledged. Any
 # commercial use must be negotiated with the Office of Technology Transfer
 # at the California Institute of Technology.
@@ -14,17 +14,11 @@
 # or other export authority as may be required before exporting such
 # information to foreign countries or providing access to foreign persons.
 
-'''
-usage: ait-table-decode --fswtabdict config/table.yaml --tabletype targets --binfile /Users/ays/targets.bin
-
-Decodes the given FSW binary table to text.
-
-Examples:
-
-  $ ait-table-decode --fswtabdict config/table.yaml --tabletype targets --binfile /Users/ays/targets.bin
-'''
+"""Decode AIT FSW table binaries"""
 
 import argparse
+import os.path
+import sys
 
 from ait.core import log, table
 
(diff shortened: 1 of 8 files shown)


MJJoyce:
This is pending a few more changes before it's fully ready. At the very least I want to include documentation updates here. The current docs for table stuff is pretty spartan (read, crappy). I also want to add Extensions support to this but I might pull that work out into a separate ticket.

MJJoyce:
This should be good to merge now that the documentation is updated. I'm going to open a follow on ticket to address the `Extensions` support to table since it's not completely trivial (I think).

Note, the NASA AMMOS org seems to have used up its TravisCI credits so the build isn't running. Lovely. Proof that the branch is passing at the moment:

Screen Shot 2021-08-05 at 14 25 18

nttoole (changes requested):
table.py has some copy/pasted comments that mention Command instead of the correct type

nttoole:
For ait_table_decode(), would there be any value in having results dumped to stdout if out_file is not provided?

MJJoyce:
Hey @nttoole,

Thanks for catching the Commanding stuff. I should have checked further when I stumbled across the mixups in the original code but forgot to follow up on it.

In terms of `ait_table_decode`, we certainly could. Happy to cover that in a follow up ticket. Most of the "real life" examples of these tables are pretty big but that's of course project specific.

MJJoyce:
@nttoole, searched through the docstrings and fixed the Command references. Should look better now.

nttoole:
@MJJoyce Looking at [test_table](https://github.com/NASA-AMMOS/AIT-Core/blob/issue-354/ait/core/test/test_table.py), is it correctly exercising the enum-type field?

MJJoyce:
> @MJJoyce Looking at [test_table](https://github.com/NASA-AMMOS/AIT-Core/blob/issue-354/ait/core/test/test_table.py), is it correctly exercising the enum-type field?

It is. That being said, I actually intended to remove the enumeration handling entirely (which I clearly failed to do) but I'm wondering if that's not the correct decision ...

Let me think on that a bit

MJJoyce:
@nttoole,
Updates pushed for enumeration handling. This is now properly handled on both encode and decode (with options to not include enumerates on decode). Tests have been added.

Heres an example run of stuff
```
(ait-core) > $ more test.txt
1, 2, 3
4, 5, 1
4, 5, TEST_ENUM_1

(ait-core) > $ ait/core/bin/ait_table_encode.py TestTable test.txt
[ ... ]
2021-08-17T13:39:19.835 | PROGRAM  | done.

(ait-core) > $ ait/core/bin/ait_table_decode.py test.bin
[ ... ]
2021-08-17T13:39:27.668 | PROGRAM  | done.

(ait-core) > $ more test_decoded.txt
# uptype=1
# HEADER_COLUMN_TWO=2
# HEADER_COLUMN_THREE=3
4,5,TEST_ENUM_1
4,5,TEST_ENUM_1
0,0,TEST_ENUM_0

(ait-core) > $ ait/core/bin/ait_table_decode.py --raw test.bin
[ ... ]
2021-08-17T13:42:47.273 | PROGRAM  | done.

(ait-core) > $ more test_decoded.txt
# uptype=1
# HEADER_COLUMN_TWO=2
# HEADER_COLUMN_THREE=3
4,5,1
4,5,1
0,0,0
```

Note in the enumerated case that fill values will get converted to their enumerated output. A potential reason that raw option might be useful ;)

PR #361, opened 2021-08-30
Issue #360 - Fix time handling in table encode / decode

### LINKED ISSUE #360: Table encode / decode doesn't properly handle Time values
The table updates from #355 rely on ait.core.dtype for encoding and decoding individual columns of data. This works fine on Primitive data types but fails to handle Time values (Technically, it shouldn't work for command stuff either. More on that later ...). Specifically, Time64Type, and Time40Type, Time32Type expect a datetime object to be passed to encode and we don't have support for that currently.

Update the encode / decode handling so "Time Types" are properly handled. Additionally, update the docs to elaborate on what data types are supported by default.

MJJoyce:
Update table encode / decode handling so fields using Time64/40/32 are
properly passed datetime objects and decode into a well formatted
datetime string. Passed time strings must be RFC3339 formatted and will
be returned as such when decoded. Tests have been updated with time
examples.

Documentation has been updated to indicate what data types we support
and also elaborate on when users may need to extend core classes if they
want to support other data types (e.g., CMD16 and EVR16).

Resolve #360

### DIFF
--- a/ait/core/table.py
+++ b/ait/core/table.py
@@ -11,6 +11,7 @@
 # laws and regulations. User has the responsibility to obtain export licenses,
 # or other export authority as may be required before exporting such
 # information to foreign countries or providing access to foreign persons.
+import datetime
 import hashlib
 import io
 import os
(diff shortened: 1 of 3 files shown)


nttoole (approved):
lgtm

PR #381, opened 2021-10-27
Update Core package handling

### LINKED ISSUE #364: Packaging and Dev Tooling Rework
A parent ticket to track various updates and additions to our packaging and dev tooling.

### Packaging
Having all of our package configuration in `setup.py` is very out of date. At the minimum we should move stuff into `pyproject.toml` (see PEP 517 / 518) and `setup.cfg`. We should also consider `Poetry` or a similar "package manager", if only for the benefit of better dependency pinning and resolution handling.

### Dev Tooling
Our tooling is pretty lacking. Outside of a single githook there's not really much of anything here.
- Swap from `nose` to `pytest`
- Add `pre-commit` for better / easier pre-commit/push hooking pipelines
- Add `black` and `flake8` for auto-formatting and linting checks
- Add `mypy` so we can start type annotating the toolkit
- Add `tox` so we can simplify our testing / build entry points. Additionally, easier to support testing across multiple versions.

MJJoyce:
Update Core to use Poetry for package management. See their home page
for additional details on how to use it and what it gets us.

https://python-poetry.org/

Update packaging to be PEP 517/518 compliant. All configuration has been
moved to setup.cfg and pyproject.toml. Setup.py has been dropped (since
we're not using setuptools now).

Documentation has been updated and moved from the wiki to the README.
Once we merge this we should drop the wiki page since a decent chunk of
that page is now out of date.

Resolve #364


nttoole (commented):
How tied would standard development be to pyenv?  The reason I ask is because it does not seem to support CSH

README.rst
```diff
@@ -21,18 +18,124 @@ Page <http://ait-core.readthedocs.io/en/latest/project_setup.html>`__
...
+
+    poetry install
+
+As with a normal installation you will need to point `AIT_CONFIG` at the primary configuration file. You should consider saving this in your shell RC file or your virtual environment configuration files so you don't have to reset with every new shell::
+
+    export AIT_CONFIG=/path/to/ait-core/config/config.yaml
+
+Finally, you should configure `pre-commit` by running the following. This will install our pre-commit and pre-push hooks::
+
+    pre-commit install && pre-commit install -t pre-push
```
nttoole, line 55: Is pre-commit a separate tool that needs its own installation? If so it should be included in the tool list above.

nttoole: Or maybe I just needed to re-start my virtual env?

README.rst
```diff
@@ -21,18 +18,124 @@ Page <http://ait-core.readthedocs.io/en/latest/project_setup.html>`__
...
+
+Individual calls to the tools are configured in ``.pre-commit-config.yaml``. If you'd like to run a specific tool on its own you can see how we call them there.
+
+
+Documentation
+~~~~~~~~~~~~~
+
+AIT uses Sphinx to build its documentation. You can build the documentation with::
+
+    poetry run build_sphinx
```
nttoole, line 86: This command resulted in error for me

```
[venv] MT-202661{nttoole}% poetry run build_sphinx

  FileNotFoundError

  [Errno 2] No such file or directory: b'/Applications/CMake.app/Contents/bin/build_sphinx'

  at /Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/os.py:605 in _execvpe
       601│         path_list = map(fsencode, path_list)
       602│     for dir in path_list:
       603│         fullname = path.join(dir, file)
       604│         try:
    →  605│             exec_func(fullname, *argrest)
       606│         except (FileNotFoundError, NotADirectoryError) as e:
       607│             last_exc = e
       608│         except OSError as e:
       609│             last_exc = e
```

nttoole:
Poetry build itself looks like its working, though there are some issues with running commands that once worked with the older build system.
Since @MJJoyce is out on temporary leave, our plan is to merge this into master and then quickly create a new PR that resolves some of the issues #383

nttoole (approved):
Approved for changes to Poetry build, tho issues will be addressed in separate Issue #383

PR #390, opened 2021-12-29
Add try except when attempting to pack opcode into struct.

Mejiro-McQueen:
This is intended to help the end user diagnose typos in their command dictionaries.

The default error message is not descriptive enough for the user to quickly find their error.
Enables concise and useful failures logs for smoke testing dictionaries.


Mejiro-McQueen:
This change prevents frustrating the user when an invalid opcode is provided in the dictionary; They'll be able to pinpoint the error immediately instead of having to probe around.

# Using a simple test case:
image

# Useful Error Reporting captured by Tests with new behavior
image


# Less useful Reporting captured by Tests with current behavior
image

PR #391, opened 2022-01-05
Add DSN paths to ait-create-dirs

Mejiro-McQueen:
Add capability for ait-create-dirs to create the directories specified by the DSN blocks in config.yaml
 
Add DSN config.yaml path entries to AitConfig._datapaths
ait-create-dirs will now be able to create directories for the DSN block.


Mejiro-McQueen:
Changing branch name

PR #393, opened 2022-01-05
Issue-392 Append 'extensions' map to flattened config map.

Mejiro-McQueen:
Fixes bug where the extensions map would be deleted from config map.
This prevented extensions from instantiating.


Mejiro-McQueen:
# Before
image

# After 
image

PR #396, opened 2022-01-05
Issue-394 Add DSN paths to ait-create-dirs

Mejiro-McQueen:
Add capability for ait-create-dirs to create the directories specified by the DSN blocks in config.yaml

Add DSN config.yaml path entries to AitConfig._datapaths
ait-create-dirs will now be able to create directories for the DSN block.

### DIFF
--- a/ait/core/cfg.py
+++ b/ait/core/cfg.py
@@ -332,6 +332,14 @@ class AitConfig(object):
             data = self._config["data"]
             for k in data:
                 paths[k] = data[k]["path"]
+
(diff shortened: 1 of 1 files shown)


nttoole (changes requested):
The code change is approved, tho we would like to retain the original config.yaml paths for the time-being.  Could the yaml updates be reverted?

Mejiro-McQueen:
Reverted config.yaml 

I don't agree with requiring a directory on / or having air-create-dirs ask for sudo. 

Maybe the user guide should suggest making the /gds/ directory with appropriate permissions.

nttoole (approved):

PR #400, opened 2022-01-11
issue-#399 Fix error when IPV6 is disabled at kernel level

### LINKED ISSUE #399: builtins.OSError: [Errno 97] Address family not supported by protocol when ipv6 disabled at kernel level
When IPV6 is disabled at the kernel level, the following error appears during runtime.
builtins.OSError: [Errno 97] Address family not supported by protocol

AMMOS Kickstarts and AMIs have disabled IPV6 at the kernel level. 
I strongly recommend against this, but alas, here's a fix.

Mejiro-McQueen:
Fix to error:

builtins.OSError: [Errno 97] Address family not supported by protocol
when ipv6 disabled at kernel level

Fixes #399

### DIFF
--- a/ait/core/server/client.py
+++ b/ait/core/server/client.py
@@ -32,7 +32,8 @@ class ZMQClient(object):
         # open PUB socket & connect to broker
         self.pub = self.context.socket(zmq.PUB)
         self.pub.connect(zmq_proxy_xsub_url.replace("*", "localhost"))
-
+        if 'listener' in kwargs and isinstance(kwargs['listener'], int) :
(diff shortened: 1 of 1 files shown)


nttoole (changes requested):
@Mejiro-McQueen Could you please include a check for the if-statement to ensure the original `listener` type is 'int'?   (I know a subclass already performs a check, but think it would be prudent to include it here).  Thanks.

Mejiro-McQueen:
Changed as requested.

nttoole (approved):

PR #416, opened 2022-02-28
testing nested includes during test for 'dirty' (pickle)

### LINKED ISSUE #371: Consider removing pickle caching of configuration files
The majority of our core configuration files are cached as pickle files when they're loaded. These files, especially thanks to the somewhat inconsistent handling of nested / included telemetry files forcing a cache reload, cause a non-trivial amount of issues for our users. One of our users runs a cronjob every ~5 minutes to delete any `*.pkl` file in the deployed AIT directories. This caching saves at most a few seconds of load time on bootup.

We should evaluate the impact of removing this caching for bootup times. It would be useful to throw some tests together to determine how slow bootup gets with "very large" dictionaries. If we find we don't care about the slowdown, we can remove them entirely. If we find that there's some potential value there, we should consider alternative solutions to avoid the out of date cache issues.

JimHofman:
This change request is based on issue-371. Users were having problems with the serialization of their config files. It was found that the included config files were not being tested for creation date, so when an included config file was modified, the need to recreate the binary (.pkl file) was not being recognized by the software. This resulted in and out of date config file being loaded. This has now been fixed with the addition of a short method (check_yaml_timestamps()) and a change to the 'dirty' property, which, now calls check_yaml_timestamps() in the ObjectCache class in util.py.

Fixes #371

### DIFF
--- a/ait/core/util.py
+++ b/ait/core/util.py
@@ -127,11 +127,11 @@ class ObjectCache(object):
             if self.dirty:
                 self._dict = self._loader(self.filename)
                 self.cache()
-                print('FILES MODIFIED - MAKE NEW PICKLE CACHE')
+                ait.log.debug('FILES MODIFIED - MAKE NEW PICKLE CACHE')
             else:
                 with open(self.cachename, "rb") as stream:
                     self._dict = pickle.load(stream)
-                    print('LOAD PICKLE CACHE.')
+                    ait.log.debug('LOAD PICKLE CACHE.')
 
         return self._dict
 
--- a/ait/core/util.py
+++ b/ait/core/util.py
@@ -59,10 +59,8 @@ class ObjectCache(object):
 
     @property
     def dirty(self):
-        """True if the cache needs to be updated, False otherwise"""
-        return not os.path.exists(self.cachename) or (
-            os.path.getmtime(self.filename) > os.path.getmtime(self.cachename)
-        )
+        """True if the pickle cache needs to be updated, False to use pickle binary"""
+        return self.check_yaml_timestamps(self.filename, self.cachename)
 
     @property
     def filename(self):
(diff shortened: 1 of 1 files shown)


ait/core/util.py
```diff
@@ -76,19 +74,64 @@ def cache(self):
...
 
         Loads the Python object, either via loader(filename) or the
         pickled cache file, whichever was modified most recently.
         """
+
         if self._dict is None:
             if self.dirty:
                 self._dict = self._loader(self.filename)
                 self.cache()
+                print('FILES MODIFIED - MAKE NEW PICKLE CACHE')
```
nttoole, line 130: Could you replace all calls of print() with ait.log.debug()

nttoole:
@JimHofman Could you add some documentation to the PR above, including what issue is being addressed and how it was resolved?  Thanks

ait/core/util.py
JimHofman: This change request is based on issue-371.  Users were having problems with the serialization of their config files.  It was found that the included config files were not being tested for creation date, so when an included config file was modified, the need to recreate the binary (.pkl file) was not being recognized by the software.  This resulted in and out of date config file being loaded.  This has now been fixed with the addition of a short method (check_yaml_timestamps()) and a change to the 'dirty' property, which, now calls check_yaml_timestamps() in the ObjectCache class in util.py.

ait/core/util.py
```diff
@@ -76,19 +74,64 @@ def cache(self):
...
+
         if self._dict is None:
             if self.dirty:
                 self._dict = self._loader(self.filename)
                 self.cache()
+                ait.log.debug('FILES MODIFIED - MAKE NEW PICKLE CACHE')
             else:
                 with open(self.cachename, "rb") as stream:
                     self._dict = pickle.load(stream)
+                    ait.log.debug('LOAD PICKLE CACHE.')
```
nttoole, line 134: Please remove this debug as its the common case.

ait/core/util.py
```diff
@@ -76,19 +74,64 @@ def cache(self):
...
 
         Loads the Python object, either via loader(filename) or the
         pickled cache file, whichever was modified most recently.
         """
+
         if self._dict is None:
             if self.dirty:
                 self._dict = self._loader(self.filename)
                 self.cache()
+                ait.log.debug('FILES MODIFIED - MAKE NEW PICKLE CACHE')
```
nttoole, line 130: Could you update so its not all caps, and also consider including the name of the affected cache/filename.

nttoole (approved):

PR #418, opened 2022-03-01
Add Packet Accumulator

Mejiro-McQueen:
Add Packet Accumulator plugin

### DIFF
--- a/ait/core/server/plugins/PacketAccumulator.py
+++ b/ait/core/server/plugins/PacketAccumulator.py
@@ -1,0 +1,51 @@
+from ait.core.server.plugins import Plugin
+from gevent import Greenlet, sleep
+
+
+class PacketAccumulator(Plugin):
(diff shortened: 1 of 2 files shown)


ait/core/server/plugins/PacketAccumulator.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def periodic_check(self):
+        while True:
+            sleep(self.timer_seconds)
+            self.emit()
+
+    def process(self, data, topic=None):
+        data_len = len(data)
+        # Does not fit, need to emit
+        if self.size_packet_queue_octets + data_len > self.max_size_octets:
```
nttoole, line 38: Do we need to append the new data after the emit() call?

Mejiro-McQueen: Good catch

nttoole (approved):

PR #426, opened 2022-03-24
Issue 358: OpenMCT Plugin support for multiple clients and filtered telemetry

### LINKED ISSUE #357: MCT Serve realtime subscriptions should manage which channels are sent to each client <Cubesat Issue>
The MCT adaptation for AIT-Core should manage connected clients with a dictionary of subscribed channels to minimize data throughput. This will allow one data source (the MCT server adapter) to receive all telemetry, but send minimal telemetry to each client, minimizing network usage.

### LINKED ISSUE #358: MCT Serve limits to one connected client <cubesat issue>
MCT Server adaptation is currently limited to one connected client (serve is a while connected loop). This should be handled with threads and callbacks to manage a dictionary of connected clients.

### LINKED ISSUE #359: MCT Serve does not accept close requests <Cubesat Issue>
Due to connecting to a client in a while loop, websocket close requests are improperly handled by the current MCT Server adaptation. This can be tested without MCT by opening a websocket connection, then attempting to close it.

nttoole:
Refactored the plugin to maintain a set of active-websockets to which realtime telemetry will be sent.
Server also maintains state of web-socket by handling messages: subscribe, unsubscribe, close
Each web-socket maintains a set of subscribed telemetry fields, so only those fields will be transmitted to the web-socket.

We also refactored how the AIT-OpenMct extensions are integrated with OpenMCT.  No longer do we download OpenMCT and modify its internal files.  Instead, we offer an example web-server that imports openmct as a dependency. This removes an earlier requirement of using an older version of OpenMCT (before they adopted Webpack).

Finally, updated documentation to capture the new approach for deployment.

Testing involved an AIT Config YAML with the following plugin info:
```
          - plugin:
              name: ait.core.server.plugins.openmct.AITOpenMctPlugin
              inputs:
                - telem_stream
              service_port: 8082
              debug_enabled: True
              database_enabled: False
```

Testing notes particular to each issue are captured in the issue.

Fixes #357 
Fixes #358
Fixes #359

### DIFF
--- a/ait/core/server/plugins/openmct.py
+++ b/ait/core/server/plugins/openmct.py
@@ -47,32 +47,70 @@ from ait.core.server.plugin import Plugin
 
 
 class ManagedWebSocket():
-    idCounter = 0  # to assign unique ids
-    PACKET_ID_WILDCARD = "*"
-    '''
+    """
     A data structure to maintain state for OpenMCT websockets
-    '''
+    """
+    idCounter = 0  # to assign unique ids
+
+    PACKET_ID_WILDCARD = "*"
+
 
     def __init__(self, web_socket, client_ip=None):
         self.web_socket = web_socket
         self.client_ip = client_ip
-        self.subscribed_set = set()
+        self._subscribed_dict = dict() #Dict from packetId to list of fieldIds
         self.is_closed = False
         self.is_error = False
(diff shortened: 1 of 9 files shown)


jasonmlkang (approved):
Tried running ait-server with openmct example server. LGTM!

PR #432, opened 2022-04-04
Issue #371 - Refresh pkl cache files when out-of-date

JimHofman:
Refactored `check_yaml_timestamp` out of `ObjectCache` and `FSWTabDictCache` and made a standalone function in _util.py_.  

Also refactored cache() (in util.py) and update() (in table.py) from the same classes respectively, and made a standalone in _util.py_ named `update_cache()`.   

Added a 3rd argument to `update_cache()` - the object to be pickled if needed.

Added log.info() call to record what it being loaded based on changes to yaml files.  

Made some changes to variable names to standarize the usage in util.py and in table.py. 

Changed a few docstrings (first line in only """ and last line is only """ less it's a one line docstring.


ait/core/table.py
```diff
@@ -456,11 +456,57 @@ def __init__(self, filename=None):
         self.fswtabdict = None
 
     def dirty(self):
-        return not os.path.exists(self.pcklname) or os.path.getmtime(
-            self.filename
-        ) > os.path.getmtime(self.pcklname)
+        """True if the pickle cache needs to be updated, False to use pickle binary"""
+        return self.check_yaml_timestamps(self.filename, self.cachename)
+
+    def check_yaml_timestamps(self, yaml_file_name, cache_name):
```
nttoole, line 462: Does this function already exist under util.py?  Can't table.dirty just call that?

ait/core/util.py
```diff
@@ -60,6 +61,7 @@ def cachename(self):
     @property
     def dirty(self):
         """True if the pickle cache needs to be updated, False to use pickle binary"""
+        print('TESTING DIRTY')
```
nttoole, line 64: Please remove debug line

ait/core/api.py
```diff
@@ -91,7 +91,8 @@ def __str__(self):
 
     @property
     def msg(self):
-        s = 'FalseWaitError: "False" boolean passed as argument to wait. Ensure wait condition args are surounded by lambda or " "'
+        s = f'FalseWaitError: "False" boolean passed as argument to wait. Ensure wait ' \
```
nttoole, line 94: Our linter now complains when using format-strings without variable-substibution taking place, so could you please remove the 'f's?

ait/core/util.py
```diff
@@ -59,77 +60,27 @@ def cachename(self):
...
+        Loads the Python object, either via loader (filename) or the
         pickled cache file, whichever was modified most recently.
         """
 
         if self._dict is None:
             if self.dirty:
                 self._dict = self._loader(self.filename)
-                self.cache()
+                update_cache(self.filename, self.cachename, self._dict)
+                log.info(f'New pickle file: {self.filename.split(".")[0]}.pkl loaded')
```
nttoole, line 83: Same as previous self.cachename comment

Issue #427, opened 2021-08-05
MCT Adaptation Dictionary issues <Cubesat issues>


mudinthewater:
1. Dictionary should be organizer. Current dictionary is a flat list, which cases performance and UI issues if you have large numbers of channels in MCT (10k+ typically)
2. Plugin for dictionary should be extensible and able to add more sources by config. IE for some cubesats you'll have cubesat dictionary 1, cubesat dictionary 2, cubesat ancillary data (ground station etc) 1, cubesat ancillary data 2.

PR #433, opened 2022-04-04
Issue 427: Update AIT OpenMCT comp. provider to serve packets as fold…

nttoole:
…ers of fields

Instead of display all telemetry fields as a flat list in OpenMCT, the changes associated with this PR organize the telemetry such that packets are folders containing underlying telemetry fields.

-----------------------------

Old (flat list):
image

-----------------------------

New (nested) with option to use full field name (packet+field):
image

-----------------------------

New (nest) with option (field name only)
image

Fixes part 1 of #427 (uncertain second part, multiple dictionaries, will be addressed until AIT supports it directly)

### DIFF
--- a/openmct/ait_integration.js
+++ b/openmct/ait_integration.js
@@ -48,24 +48,42 @@
  *  3) Point your browser to localhost:8080
  */
 
-//DEFAULTS
+
+
+//-----
+//Constants
+
+let OBJ_NAMESPACE = "ait-ns";
+let OBJ_ROOT = "tlm-dict";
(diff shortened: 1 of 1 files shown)


jasonmlkang (approved):
LGTM! Only question is do you also need to update `ait_integration.js` under `example-server` as well?

nttoole:
> LGTM! Only question is do you also need to update `ait_integration.js` under `example-server` as well?

That file is a symlink to the original, so we should be ok.  Thanks.

PR #447, opened 2022-06-08
Issue #370: Use TempFile class throughout test suite

### LINKED ISSUE #370: Use TempFile class throughout our test suite
We have a [`TestFile` class](https://github.com/NASA-AMMOS/AIT-Core/blob/40abda33b2a32a49eeb4d00d373f3b49aa1e7981/tests/ait/core/__init__.py#L44) that provides a lightweight abstraction around `tempfile`. Some of our tests use this but many don't. We should update all our tests to use this abstraction. 

We should also consider moving this class out of [ait.core.__init__](https://github.com/NASA-AMMOS/AIT-Core/blob/master/ait/core/test/__init__.py#L47) into something a bit more obvious. Maybe `ait.core.util`?

JimHofman:
In unit tests removed all code using _'tempfile'_ and replaced with code using the _TestFile_ class.
Moved _TestFile_ class from _tests/ait/core/__init__.py_ to _ait/core/util.py_.
Modified _TestFile_ class to include on _'options'_ argument for file creation options (e.g. "w+", "wb+",...)
Factored out redundante _TestFile_ class in _test/ait/core/server/test_apid_rounting.py_ and used the common class in _util.py_
Did some minor clean up of import orders, 1 change to a class name, and a couple of docstrings to more closely match PEP 8 style guide.
Verified no _'tempfile'_ imports exist in the _'tests'_ directory structure
Ran all the unittests

Updated the syntax for calling the Context Manager to match docstrings.

Fixes #370

### DIFF
--- a/ait/core/table.py
+++ b/ait/core/table.py
@@ -16,9 +16,10 @@ import hashlib
 import io
 import os
 import pickle
+import yaml
 
 import ait
-import yaml
+
 from ait.core import dmc
 from ait.core import dtype
 from ait.core import log
--- a/ait/core/util.py
+++ b/ait/core/util.py
@@ -26,6 +26,8 @@ import stat
 import sys
 import time
 import zlib
+import tempfile
+import warnings
 
 import pickle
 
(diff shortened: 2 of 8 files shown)


nttoole (changes requested):
There seems to be a lot of instances of:
```
        test_file = TestFile("test_table_in.bin", "wb+")
        with test_file as temp_file:
```
...that should probably be streamlined to:

`        with TestFile("test_table_in.bin", "wb+") as temp_file:`

nttoole (approved):

Issue #450, opened 2022-07-07
Create plugin which keeps track of APID-specific sequence counts


kmarwah:
Create a plugin which maintains a dictionary of the current sequence count for each APID and raises a warning if there are missing packets. The dictionary can be stored as a .pkl file so that the count persists across sessions.

PR #451, opened 2022-07-07
Issue #450 : track sequence count of received CCSDS packets

kmarwah:
Implements a plugin which checks the sequence count field of received CCSDS packets. A warning message is logged if an unexpected sequence count value is received. The plugin maintains a dictionary of the expected sequence count values for each APID, and this dictionary is stored as a .pkl file.

### DIFF
--- a/ait/core/server/plugins/apid_sequence_count.py
+++ b/ait/core/server/plugins/apid_sequence_count.py
@@ -1,0 +1,143 @@
+'''
+Implements a plugin which maintains a sequence count for each APID
+'''
+import os
+import pickle
+import ait
+import ait.core
+from ait.core.server.plugins import Plugin
+from ait.core import log
+
+
+class APIDSequenceCounter(Plugin):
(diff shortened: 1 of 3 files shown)


ait/core/server/plugins/apid_sequence_count.py
```diff
@@ -0,0 +1,143 @@
...
+class APIDSequenceCounter(Plugin):
+    '''
+    Checks the sequence count field of CCSDS packets to ensure that packets aren't dropped.
+    Maintains a dictionary of the expected sequence count value for each APID.
+    The dictionary is saved and loaded as .pkl files so that the count persists across sessions.
+
+    example setup in config.yaml:
+
+    - plugin:
+        name: ait.core.server.plugins.apid_sequence_path.APIDSequenceCounter
```
nttoole, line 21: @kmarwah I believe there is a type in the full name?

PR #452, opened 2022-07-07
Issue #201: Yaml file schema verification now works on included yaml files

### LINKED ISSUE #201: YAML schema validation doesn't check nested files
The current YAML schema validation does not check nested YAML files from `includes`. We'll need to update `YAMLProcessor` and `Validator` in order to add this functionality.

JimHofman:
Issue-201 Yaml file schema verification now works on included yaml files.

In the Validator class I renamed the validate() method validate_schema().  This is used in the unit test to validate evr.yaml, table.yaml and limits.yaml.  It is only  a  general schema check.

I removed the individual validate() methods from the CmdValidator(Validator) and TlmValidator(Validator) classes and moved a general validate() function to the Validator class.

Added get_included_files() to the Validator class to produce a list of yaml files that need to be validated.

Fixes #201

### DIFF
--- a/ait/core/cmd.py
+++ b/ait/core/cmd.py
@@ -558,8 +558,6 @@ def YAMLCtor_include(loader, node):  # noqa
     # Get the path out of the yaml file
     name = os.path.join(os.path.dirname(loader.name), node.value)
     data = None
-    # DEBUG
-    log.debug(f'NAME of File: {name}')
     with open(name, "r") as f:
         data = yaml.load(f)
     return data
--- a/ait/core/cmd.py
+++ b/ait/core/cmd.py
@@ -526,7 +526,8 @@ def getMaxCmdSize():  # noqa
 
 
 def handle_includes(defns):
-    """Recursive handling of includes for any input list of defns.
+    """
+    Recursive handling of includes for any input list of defns.
     The assumption here is that when an include is handled by the
     pyyaml reader, it adds them as a list, which is stands apart from the rest
     of the expected YAML definitions.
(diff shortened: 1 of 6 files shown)


ait/core/cmd.py
```diff
@@ -557,6 +558,8 @@ def YAMLCtor_include(loader, node):  # noqa
     # Get the path out of the yaml file
     name = os.path.join(os.path.dirname(loader.name), node.value)
     data = None
+    # DEBUG
+    log.debug(f'NAME of File: {name}')
```
nttoole, line 562: Could this either be removed, or make the debug statement more meaningful?  Seeing it in a log as is wouldn't be too helpful.

ait/core/val.py
```diff
@@ -360,30 +359,118 @@ def pretty(self, start, end, e, messages=None):
...
+                            self.validate_list.append(f'{self.yml_dir}/{included_file_name}')
+                            # Look for includes within included file
+                            self.get_included_files(included_file_name)
+                    # Check and flag multiple includes
+                    if len(self.validate_list) != len(set(self.validate_list)):
+                        log.info(f'WARNING: Validate: Duplicate config files in the '
+                                 f'include tree. {self.validate_list}')
+                    return set(self.validate_list)
+            except RecursionError as e:
+                print(f'ERROR: {e}: Infinite loop: check that yaml config files are not looping '
```
nttoole, line 425: Switch to logger call

JimHofman: Done, also made the same change in util.py, where there is similar code.

nttoole (approved):

PR #455, opened 2022-07-20
Issue #445 - Add multiprocess support for Plugins

### LINKED ISSUE #445: Add Plugin class that runs as process instead of greenlet
SunRise reported performance issues with their architecture which relies on Plugins for the workflow.
On the downlink side, there is a single Plugin.
On the uplink side, there are multiple (eight?) Plugins.

When running with downlink and uplink, there is a load of commands being pushed on the later.
This results in the downlink performance dropping significantly, to the point that buffers fill and timeouts are reached.

One possible way of dealing with this is to refactor the Server Plugin (which currently extends greenlet), into a set of Plugin classes, one that continues to extend greenlet, and one that runs as a process.
Communication via 0MQ _should_ makes this relatively painless, tho there are still details of process lifecycle and potential issues coordinating with greenlets.

nttoole:
This PR adds support for running Plugins in separate processes, as opposed to the current behavior of all Plugins running in the server process.

We maintained the Plugin API such that it remains an extension of a greenlet, so no existing plugins require any updates.

We made it relatively simple to configure a plugin to take advantage of multiprocessing.  There is a new optional attribute `process_id`, which if defined, will associated that Plugin with its own process.  

```
    server:
        plugins:
            - plugin:
                process_id: ait_gui
                name: ait.gui.AITGUIPlugin
                inputs:
                    - log_stream
                    - telem_stream
                outputs:
                    - command_stream
```
If multiple plugins declare the same value for `process_id`, then all of those respective plugins will run in that one process.

The server now maintains separate lists of plugins and plugin-processes.  Plugin-processes collect Plugin info's (an abstraction of the Plugin config), and once spawned, create the Plugin instances _within_ that process space.  We utilize the [GIPC library](https://gehrcke.de/gipc/) to ensure a cleaned gevent space per process.  

Since we already leverage 0MQ for server communication, the only difference for plugins running in a separate process is that they are setup using a fresh 0MQ context (as opposed to reusing the Broker's context).

We also added [setproctitle](https://pypi.org/project/setproctitle/) as a dependency to allow server to rename the process title to include the process_id/namespace.  This process_id will appear as part of the process when listing via top/ps.

```
MT-202661{nttoole}% ps -eff | grep ait
1524906997  5233 10872   0  4:32PM ttys001    0:00.64 /Library/Frameworks/Python.framework/Versions/3.7/Resources/Python.app/Contents/MacOS/Python ./ait/core/bin/ait_server.py
1524906997  5239  5233   0  4:32PM ttys001    0:00.09 /Library/Frameworks/Python.framework/Versions/3.7/Resources/Python.app/Contents/MacOS/Python ./ait/core/bin/ait_server.py plugin-process.ait_gui 
```
In the example above, the first process is the AIT-Server, and the second process is solely the AITGUIPlugin.

For testing, we ran the through standard GUI setup case, but instead had the GUI in its own process.  We demonstrated that telemetry would flow from the server to the GUI plugin, as well as commands from plugin to the server.

Lastly, documentation for the server was updated.

Fixes #445


jgehrcke:
Lovely use of gipc. Thanks!

PR #456, opened 2022-10-06
Issue 368

JimHofman:
Add error handling for PacketHandler and CCSDSPacketHandler for check on the size relative tot he configured packet definitions.

### DIFF
--- a/ait/core/server/handlers/packet_handler.py
+++ b/ait/core/server/handlers/packet_handler.py
@@ -19,27 +19,48 @@ class PacketHandler(Handler):
                            If packet is specified but not present in default tlm dict.
         """
         super(PacketHandler, self).__init__(input_type, output_type)
-        self.packet = kwargs.get("packet", None)
+        self.packet_type = kwargs.get("packet", None)
+        self.tlm_dict = tlm.getDefaultDict()
 
-        if not self.packet:
-            msg = 'PacketHandler: No packet name provided in handler config as key "packet"'
+        if not self.packet_type:
+            msg = f'PacketHandler: No packet type provided in handler config as key "packet"'
             raise ValueError(msg)
 
-        tlm_dict = tlm.getDefaultDict()
-        if self.packet not in tlm_dict:
-            msg = "PacketHandler: Packet name {} not present in telemetry dictionary".format(
-                self.packet
-            )
-            msg += " Available packet types are {}".format(tlm_dict.keys())
+        if self.packet_type not in self.tlm_dict:
+            msg = f"PacketHandler: Packet name '{self.packet_type}' not present in telemetry dictionary."
+            msg += f" Available packet types are {self.tlm_dict.keys()}"
             raise ValueError(msg)
 
-        self._pkt_defn = tlm_dict[self.packet]
+        self._pkt_defn = self.tlm_dict[self.packet_type]
 
-    def handle(self, input_data):
+    def get_packet_lengths(self):
+        """
+        Makes a dictionary of packet.name : number of bytes in the packet
+            e.g.  'Ethernet_HS_Packet': 37
+
+        Return:  dictionary
+
+        """
+        pkt_len_dict = {}
+        for i in self.tlm_dict.keys():
+            pkt_len_dict[i] = self.tlm_dict[i].nbytes
+
+        return pkt_len_dict
+
+    def handle(self, packet):
         """
         Params:
-            input_data:   message received by stream
+            packet:   message received by stream (packet)
         Returns:
             tuple of packet UID and message received by stream
         """
-        return pickle.dumps((self._pkt_defn.uid, input_data), 2)
+
+        # TODO validate the packet
+
+        defined_packet_lengths = self.get_packet_lengths()
+
+        if defined_packet_lengths[self.packet_type] != packet.nbytes:
+            msg = f"PacketHandler: Packet length of packet does not match packet definition."
+            raise ValueError(msg)
+
+        return pickle.dumps((self._pkt_defn.uid, packet), 2)
--- a/ait/core/server/handlers/packet_handler.py
+++ b/ait/core/server/handlers/packet_handler.py
@@ -55,7 +55,7 @@ class PacketHandler(Handler):
             tuple of packet UID and message received by stream
         """
 
-        # TODO validate the packet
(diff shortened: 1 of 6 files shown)


ait/core/server/handlers/packet_handler.py
```diff
@@ -19,27 +19,48 @@ def __init__(self, input_type=None, output_type=None, **kwargs):
...
-                self.packet
-            )
-            msg += " Available packet types are {}".format(tlm_dict.keys())
+        if self.packet_type not in self.tlm_dict:
+            msg = f"PacketHandler: Packet name '{self.packet_type}' not present in telemetry dictionary."
+            msg += f" Available packet types are {self.tlm_dict.keys()}"
             raise ValueError(msg)
 
-        self._pkt_defn = tlm_dict[self.packet]
+        self._pkt_defn = self.tlm_dict[self.packet_type]
```
nttoole, line 34: @JHofman728 Since you are already grabbing the packet definition here, there really should be no need for the new `get_packet_lengths()` function below.

ait/core/server/handlers/packet_handler.py
```diff
@@ -19,27 +19,48 @@ def __init__(self, input_type=None, output_type=None, **kwargs):
                            If packet is specified but not present in default tlm dict.
         """
         super(PacketHandler, self).__init__(input_type, output_type)
-        self.packet = kwargs.get("packet", None)
+        self.packet_type = kwargs.get("packet", None)
+        self.tlm_dict = tlm.getDefaultDict()
 
-        if not self.packet:
-            msg = 'PacketHandler: No packet name provided in handler config as key "packet"'
+        if not self.packet_type:
```
nttoole, line 25: Could you please update this field to be `packet_name`?

ait/core/server/handlers/packet_handler.py
```diff
@@ -19,27 +19,48 @@ def __init__(self, input_type=None, output_type=None, **kwargs):
...
         Returns:
             tuple of packet UID and message received by stream
         """
-        return pickle.dumps((self._pkt_defn.uid, input_data), 2)
+
+        # TODO validate the packet (if this is the place to do the validation)
+
+        defined_packet_lengths = self.get_packet_lengths()
+
+        if defined_packet_lengths[self.packet_type] != packet.nbytes:
```
nttoole, line 62: As mentioned above, you only ever be comparing to self._pkt_defn
So the statement should look more like:
`if self._pkt_defn.nbytes != packet.nbytes:`

ait/core/server/handlers/packet_handler.py
```diff
@@ -19,27 +19,30 @@ def __init__(self, input_type=None, output_type=None, **kwargs):
...
         Params:
-            input_data:   message received by stream
+            packet:   message received by stream (packet)
         Returns:
             tuple of packet UID and message received by stream
         """
-        return pickle.dumps((self._pkt_defn.uid, input_data), 2)
+
+        if self._pkt_defn.nbytes != packet.nbytes:
+            msg = f"PacketHandler: Packet length of packet does not match packet definition."
```
nttoole, line 45: Could you please rephrase to: "PacketHandler: Packet data length does not match packet definition"

nttoole (dismissed):

nttoole:
@JimHofman Were you able to successfully run pytest on this branch?
I am getting errors like:
`_pickle.PicklingError: Can't pickle <class 'ait.core.tlm.PacketDefinition'>: it's not the same object as ait.core.tlm.PacketDefinition`
...which I _presume_ is related to one of apparently **thousands of environment issues** on my system, but wanted to confirm.

ait/core/server/handlers/packet_handler.py
```diff
@@ -19,27 +19,30 @@ def __init__(self, input_type=None, output_type=None, **kwargs):
...
-            input_data:   message received by stream
+            packet:   message received by stream (packet)
         Returns:
             tuple of packet UID and message received by stream
         """
-        return pickle.dumps((self._pkt_defn.uid, input_data), 2)
+
+        if self._pkt_defn.nbytes != packet.nbytes:
+            msg = f"PacketHandler: Packet data length does not match packet definition."
+            raise ValueError(msg)
```
nttoole, line 46: @JimHofman Could this be an error log message instead of thrown error (same as CCSDS handler)

JimHofman: Done, will work on pickle error before pushing.

JimHofman: Looks like the pickle error is coming from the unit tests that test bad packet length.  It's is set up by passing a 1552 packet to an Ethernet handler, so it makes sense.

tests/ait/core/server/test_handler.py
```diff
@@ -62,6 +67,59 @@ def test_execute_handler_returns_handle_return_on_input(self, handle_mock):
...
+        e_packet = tlm.Packet(self.ethernet_pkt, self.pkt_data)
+        result = self.handler.handle(e_packet)
+        assert 'Ethernet Health and Status Packet' in str(result)
+
+    # Send a 1553 packet to an Ethernet_HS_Packet Handler
+    def test_bad_packet_length(self):
+        pkt_1553 = self.tlm_dict['1553_HS_Packet']
+        packet = tlm.Packet(pkt_1553, self.pkt_data)
+        with pytest.raises(ValueError):
+            self.handler.handle(packet)
```
MJJoyce, line 120: `handle()` should not be getting called with a `tlm.Packet` object as the input. It should be receiving raw bytes. The expectation on these handlers is they're classifying raw data as a particular packet type. If you already have a packet why in the world would you be doing that?

The docstring on the handle functions even specifies this

```
 """
        Params:
            input_data:   message received by stream
        Returns:
            tuple of packet UID and message received by stream
        """
 ```

ait/core/server/handlers/packet_handler.py
MJJoyce: This check doesn't even look valid? You should be checking the length of the received input data not the number of bytes in a packet definition.

MJJoyce (changes requested):

tests/ait/core/server/test_handler.py
JimHofman: ok - I did not realize 'message received by stream' meant 'raw bytes'.

Issue #459, opened 2022-11-14
Python 3.10 Support


christianmkuss:
Related to #363

Currently AIT-Core is pinned to 3.7, which appears to come from several of the dependencies, as seen when trying to build the project for Python 3.10 and failing to build wheels for the dependencies. The dependencies could probably be updated to their 3.10 versions and imports fixed in order to add support for Python 3.10. This would probably drop support for Python 3.7 and potentially skip 3.8 & 3.9.

From first investigation the versions that will need to be updated are:
- bottle 0.12.17 -> 0.12.23
- greenlet 0.4.16 -> 1.1.3
- gevent 1.4.0 -> 21.12.0
- pyzmq 18.1.0 -> 24.0.0

MJJoyce:
Hey @christianmkuss thanks for opening this up and starting to poke around. We definitely want to get updates going on this. It's very likely we'll need to care about the 3.8 and 3.9 support due to various project needs. Dropping 3.7 isn't too big of a concern given the EOL coming up.

If you decide to poke at this let us know! We're always happy to have contributions.

PR #460, opened 2022-11-18
Issue #459 - Update to support Python3.10

christianmkuss:
Because of the strictly pinned versions this library would only work with Python 3.7, as stated in the documentation. Since the dependencies are well founded they provide support for newer versions of Python, including Python 3.10. By migrating to the newer versions this library will now support Python3.10 but drop support for other versions.


MJJoyce:
Thanks for the PR @christianmkuss! The team is getting back up to speed after being out of office last week for the holidays. We'll get this reviewed as soon as possible.

nttoole:
Using python 3.10.5, we installed and successfully ran the AIT Server, including GUI plugin (with minor mod for that repo to use 3.10)
image

nttoole (approved):

PR #463, opened 2022-12-08
Update tox with multiple python versions for testing

MJJoyce:
Update tox with test envs for Python 3.7 - 3.10. Skipped 3.11 due to some issues getting the tests and the like to cooperate. Likely doable with a bit of effort but didn't want perfect to be the enemy of good.

Semi-recent Poetry broke the way we were integrating with tox. The tests have been updated in line with the recommendation on Poetry's website. The development dependencies plugin we were using no longer works since it relied on some private APIs that have since changed. It seems like there isn't a great solution for this problem at the moment; Some of the alternatives are also having trouble with the new version of Poetry. Hopefully tox will support parsing pyproject.toml directly soon and do away with all these hacks.

It is recommended that people run tox inside of `poetry shell` without any other virtual environment open. I ran into a number of annoyances with tox not being able to find the correct python versions and the like when doing it other ways. The documentation has been updated to reflect this. While I was in there I tried to make the development tools section a bit clearer in general.

Minor updates throughout the repo are either from the linting checks actually being run on the code or small fixes necessary to make the linter happy. Shame on whomever has been skipping them :), and shame on me for not fixing the broken CI yet.

The python version specified in pyproject.toml has been opened up to support anything from Python 3.7 - 3.10. I've also unpinned gevent (since it seemed necessary to get 3.9 to cooperate to do some oddities with cffi on ARM). Dependencies have been relocked. Tox suite is all passing for me currently. That includes tests running on all the supported Python versions. We'll need more thorough testing here but hopefully this is a step in the correct direction.

I've also included .python-version with a list of the python versions I'm using in `pyenv`. Others should be able to use this to make sure we'll all testing against the same versions. Not sure how correct it is to include that but figured it wouldn't hurt for now.


nttoole (approved):

nttoole:
1. Was able to run `tox` within `poetry shell`, resulting in all envs passing:
```
  py37: commands succeeded
  py38: commands succeeded
  py39: commands succeeded
  py310: commands succeeded
  docs: commands succeeded
  lint: commands succeeded
  congratulations :)
```

2. Exited poetry-shell, and created an explicit virtual-env (using default py3.7)
From there, ran poetry-install, then set AIT_CONFIG, and was able to run `pytest` successfully:
```
platform darwin -- Python 3.7.6, pytest-7.2.0, pluggy-1.0.0
...
250 passed, 2 skipped, 1 warning in 7.17s
```

3. deactivate that venv, entered bash, and ran pyenv shell 3.9.13
Created and activated a new venv, ran poetry install, set AIT_CONFIG and ran pytest successfully
```
platform darwin -- Python 3.9.13, pytest-7.2.0, pluggy-1.0.0
...
250 passed, 2 skipped, 1 warning in 10.20s
```

Issue #376, opened 2021-10-18
Add extension capability to table and dtype


JimHofman:
Add the capability to extend Classes with table and dtype

MJJoyce:
Working on a patch for dtype extensions support. Will have a PR up shortly for review

cswoodruff:
It seems like extending CustomTypes requires python 3.10 or greater given the example code in the documentation. If so, this should be reflected in the install instructions.

```
class MyCustomTypes(dtype.CustomTypes):
    def get(self, typename: str) -> Optional[dtype.PrimitiveType]:
        match typename:
            case 'My24BitMSB':
                return My24BitMSB()

        return None
```

MJJoyce:
Hey @cswoodruff, I'm guessing you're referencing the `match` in the example? Or something else and I'm just overlooking it? There's no "requirement" there beyond overwriting the interface.

cswoodruff:
I was referring to the `match` in the example. It just caught me off guard because my virtualenv was running with python3.8 and I tried the example listed in the documentation and it failed.

MJJoyce:
> I was referring to the `match` in the example. It just caught me off guard because my virtualenv was running with python3.8 and I tried the example listed in the documentation and it failed.

Updated in 69ad7fe3abfd497e6ad989b10aa40f84137d093e

PR #468, opened 2022-12-15
Issue #376 - Add extension support to dtype.py

MJJoyce:
Update dtype.py to support user extensions via `dtype.CustomTypes`.
Calls to `dtype.get` now also check `CustomTypes.get` for class
instances. Users can extend `dtypeCustomTypes` via the normal
toolkit extensions interface.

Default `dtype.CustomTypes.get` always returns None so normal
behavior remains unchanged.

Add top-level Data Type documentation page. This page now explicitly
lists the built-in types that AIT supports. Additionally, information on
how to use the newly added extension support in dtypes is included as
well.

Update README's `sphinx-apidoc` call instructions to use
`--implicit-namespaces`. This is necessary to ensure that the `ait`
namespace is picked up and included in the API docs that are generated.
Without this files will be written starting with the `core` module. Update
API docs with latest changes.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -107,7 +107,7 @@ To view the documentation, open ``doc/build/html/index.html`` in a web browser. 
 
 If you need to update the auto-generated documentation you can run the following command to rebuild all of the package documentation::
 
-    sphinx-apidoc --separate --force --no-toc -o doc/source ait ait/core/server/test ait/core/test
+    sphinx-apidoc --separate --force --no-toc -o doc/source ait --implicit-namespaces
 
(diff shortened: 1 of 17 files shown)


nttoole (approved):

nttoole:
Approved with the understanding that there is expected planned rework with the type class hierarchy (an ABC class from which PrimitiveType, String, etc extend)

MJJoyce:
> Approved with the understanding that there is expected planned rework with the type class hierarchy (an ABC class from which PrimitiveType, String, etc extend)

:+1: I'll open a follow on ticket or two

PR #471, opened 2022-12-20
Issue #467 - Add baseline GH Actions build

### LINKED ISSUE #467: Automate testing with GH Actions
Move our test builds off of Travis and into GH actions. Initial delivery of this should handle unit testing. Further iterations should include the entire linting pipeline, documentation builds, and testing across multiple python versions via tox. Integrate these builds into our PR workflow.

MJJoyce:
Add a default build for any push to master or pull request that runs
the full `tox` build. This includes the test suite for all supported
python versions, the docs build, and the linting pipeline.

This is sufficient for a first pass but there are a couple of things we
might want to poke at. I could not get `actions/setup-python` caching of
poetry dependencies to work with `tox-gh-actions`. The lint run kept
failing because it couldn't find `pre-commit` even though it seemed to
have been installed properly. Should poke at this in a follow on ticket.

Would also be good to get a better build running to test package
deployment and the like. Bundling everything up and shipping it off to
test.pypi would be great. I'd also like to get a release build
integrated so we aren't manually doing that. Automatic CHANGELOG updates
after a push to master or PR merge would also be great.

Resolve #467

### DIFF
--- a/.github/workflows/full_build.yaml
+++ b/.github/workflows/full_build.yaml
@@ -1,0 +1,42 @@
+name: Full Test and Lint
+on:
+  push:
+    branches:
+      - 'master'
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  ait_build:
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
+    runs-on: "ubuntu-latest"
+    steps:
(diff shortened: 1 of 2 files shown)


MJJoyce:
See [here](https://github.com/MJJoyce/AIT-Core/actions/runs/3742564569) for an example build using this workflow

nttoole (approved):

PR #490, opened 2023-01-19
Issue #489 - Add PEP 561 compatibility changes

### LINKED ISSUE #489: Indicate PEP 561 typing info in package
We need to update Core to indicate that it is PEP 561 "compliant" so users can leverage type checkers with Core code.

While messing with this in GUI it seems like we need to:

- Add a `py.typed` file in the package
- Include `py.typed` in MANIFEST.in (maybe, who knows really)

MJJoyce:
Resolve #489

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,2 @@
 include ait/core/data/*.json
+include ait/core/py.typed


nttoole (dismissed):

MJJoyce:
Rebased to clear out the merge commit and updated. Merging.

Issue #507, opened 2024-01-05
EVR and Limit dictionary validation failing unexpectedly


MJJoyce:
Noted unexpected failures when verifying the default EVR and Limit files in
latest AIT Core. Zero useful error messages reported to help narrow issue down.

Snippet from logs:
```
2024-01-05T13:44:42.271 | PROGRAM  | ait-core/bin/ait-yaml-validate --evr
2024-01-05T13:44:42.278 | INFO     | Validating: AIT-Core/config/evr.yaml
2024-01-05T13:44:42.281 | ERROR    | Validation: FAILED: yml=AIT-Core/config/evr.yaml, schema=AIT-Core/ait/core/data/evr_schema.json
2024-01-05T13:44:42.282 | PROGRAM  | done.
```

PR #508, opened 2024-01-05
Issue #505 - Update hooks to run YAML verification on every commit

### LINKED ISSUE #505: Update pre-commit hooks to always run dictionary checks
Update pre-commit hooks so dictionary YAML files are validated on every commit
instead of only on push. This ensures that our default builds catch potential
issues with YAML files and helps make errors obvious early when doing related
dev work.

MJJoyce:
Resolve #505

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -45,7 +45,6 @@ repos:
         name: AIT TLM YAML check
         entry: 'ait-yaml-validate --tlm'
         language: system
-        stages: [push]
         pass_filenames: false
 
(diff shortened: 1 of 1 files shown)


MJJoyce:
Note, expecting builds to fail as a result of these changes until #507 is fixed

nttoole (approved):
Confirmed that pre_commit hook shows YAML validation error:

```
AIT LIMITS YAML check....................................................Failed
- hook id: limits_yaml_check
- exit code: 1
```

PR #513, opened 2024-01-08
Update method calls to use read_binary() and read_text().

### LINKED ISSUE #504: ait.core.seq.py has references to old names of read_text() and read_binary()
This is on AIT Core v2.5.1, see https://github.com/NASA-AMMOS/AIT-Core/blob/1fdfd45e9085c036df2a1279643efb468490f424/ait/core/seq.py#L165-L168

lockhart:
There was a lingering reference to readBinary() and readText(). This updates two lines to use read_binary() and read_text() instead.

Resolves #504

### DIFF
--- a/ait/core/seq.py
+++ b/ait/core/seq.py
@@ -163,9 +163,9 @@ class Seq(object):
         stream.close()
 
         if magic == self.magic:
-            self.readBinary(filename)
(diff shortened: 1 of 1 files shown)


MJJoyce:
Thanks for fixing this @lockhart

Issue #518, opened 2024-03-11
Update pre-commit to point to ait folder


cjjacks:
pre-commit seems out of date. for some checks, a `src` folder is referenced.  I think this should instead point to `ait`.

nttoole:
Closed by #519

PR #519, opened 2024-03-11
Issue #518 - Update pre-commit config

cjjacks:
Point reorder_python_imports and black to correct dirs


nttoole (approved):
Thank you for catching this.

PR #532, opened 2024-07-17
Issue #530 - Use SafeLoader for YAML load calls

### LINKED ISSUE #530: Multiple Local Code Executions (using YAML)
Multiple Local Code Executions (using YAML)

AIT-Core uses the YAML Python library (pyyaml) in version 5.3.1, which is vulnerable to Command Execution vulnerabilities. By crafting a malicious YAML file, the bad actors can achieve a Code Execution on the target system that runs AIT-Core. Figure 16 shows an example of a malicious YAML file containing a command that will be executed upon loading that file.

Figure 16: Example of a malicious YAML file.

To demonstrate this, we have created a simple Python script that imports the TLM module and attempts to read Telemetry definitions from a YAML file (see Figure 17).

Figure 17: Test script that loads TM definitions.

Once the script is executed, the AIT-Core TLM module loads the file, and due to YAML vulnerability, the command included in the malicious file is executed on the target system (see Figure 18).

Figure 18: Code Execution by trying to load TM definitions.

As in the previous case, AIT-Core extensively uses the YAML library. If a bad actor can access any of the YAML files used by AIT-Core or the main configuration, they can exploit the Command Execution vulnerability. Figure 19 shows a snippet of the main AIT-Core configuration file referring to other YAML files and the framework areas this vulnerability affects.

Figure 19: Other areas of concern

**Recommendations**

PyYAML offers a safe_load method that uses the SafeLoader class, which is recommended in version 6.0 and later. This method only uses standard YAML tags and avoids constructing arbitrary Python objects. Using SafeLoader ensures you’re only reading data objects composed of Python primitives and basic data structures, which is usually sufficient. If you require state and behaviour, PyYAML can derive from the YAMLObject class to create safe, user-defined types recognisable by SafeLoader.

nttoole:
The primary changes address using safer version of YAML loads:
1) swapping `yaml.load(source,Loader=yaml.Loader)` to `yaml.safe_load(source)`. (similar for `load_all`)
2) swap Yaml object constructors from `yaml.add_constructor()` to  `yaml.SafeLoader.add_constructor()`
3) update pyyaml version to '6.0.1'

Then we hit a series of pre-commit errors due to newly-corrected configs which _now_ run reorder-python-imports and black on the source code.  So to satisfy those, more non-yaml updates were made.  For example, `AitConfigMissing` was renamed `AitConfigMissingError`.
However, after a while, it became clear that black and flake8 have some contradictory rules ( e.g. 'binary logical operators across multiple lines' vs 'line-too-long' errors).  In the end, we disabled checks for the binary-ops-spanning-multi-line checks in `flake8`, cuz that just seems a cleaner presentation (to me at least).
So there are a lot of line changes related to the above, had I known I would have just avoided satisfying any pre-commit issues and forced the Yaml-related changes only.

`flake8` would still complain about source files which weren't even touched for the PR, so in the end the pre-commit was disabled so this PR could be made without touching nearly every source file (a separate PR for that busy-work).

Testing included automated tests, along with running the server with the AIT-GUI plugin and ait_example which successfully showed telemetry graphed in the browser.  (Config file was reset to clear out plugins.)

Fixes #530


MJJoyce:
Looks good to me. +1 to merge as is and resolve any linting / build issues separately.

PR #535, opened 2024-08-30
Issue #531 - Add Security notes to the server documentation

### LINKED ISSUE #531: Remote Code Execution
AIT-Core comes with a server that enables the ZMQ topic/streams to make telemetry and telecommands packets available remotely, i.e., via API script. To start a simple server, importing an Instrument from the AIT-Core API and instantiating it is enough (see Figure 20).

Figure 20: Creation of AIT-Core server.

The created instrument is now ready to subscribe to a telemetry source. For this purpose, we used the TLM module to create a Telemetry simulator, which retrieves a definition of a simple TM packet and sends it every 5 seconds (see Figure 21). In addition, to make the test more realistic, we have set up the TLM Instrument and TLM simulator on two different hosts. To exchange the telemetry, the AIT-Core opens a new port (5560 in our case), establishes the connection between the processes, and communicates using ZeroMQ messaging.

By monitoring the network traffic, we captured the TCP Frames exchanged between the instrument and the TLM simulator (see Figure 22). Given that the ZeroMQ communication used in the AIT-Core is unencrypted, we were able to capture the raw TCP Frames and analyse their content. The Python Pickle object serialisation/deserialisation library powers the TLM functionality. As per Python documentation, this library is insecure and allows bad actors to achieve Command Execution. Given that the communication is unencrypted, a bad actor can conduct a Man-in-the-Middle (MitM) attack to capture the raw TCP Frames and modify their content. With Pickle in the loop, this can lead to an RCE on either of the hosts in this communication link without access to any of them.

Figure 21: AIT-Core TM simulator.

Figure 22: Communication between the instrument and TM simulator.

We have used our tool to conduct a MitM attack to demonstrate this. First, based on the previously captured TCP Frames, we reconstructed a valid TM packet, which, when encoded, looks like the data shown in Figure 23.

Figure 23: Valid TM Packet.

Knowing that once the TLM Instrument receives this data, it will try to load it into the Pickle library, we have developed the following exploitation scenario:

1. Craft a payload that opens a reverse shell.
2. Capture a TCP Frame that contains a valid ZMQ topic and TM packet data.
3. Follow the same procedure as the TLM module to craft a new TCP Frame.
4. Pickle the reverse shell payload twice (in the original procedure, AIT-Core includes the TM Packet UID, but for exploitation, it is unnecessary).
5. Include the ZMQ topic.
6. Replace the data field of the captured TCP Frame with the maliciously crafted one.

Figure 24 shows an example of a reverse shell payload that we have decided to use for this exploitation scenario.

Figure 24: Example of a reverse shell payload.

After performing all required steps (encoding, dumping data with pickle, and adding the topic), the new TCP Frame data field looks like Figure 25.

Figure 25: TCP Frame containing a malicious payload.

Using our tool to conduct the MitM attacks at the TCP Frame level, we could replace a valid TCP Frame with one containing the reverse shell. Figure 26 shows the execution of our exploit. Figure 27 shows how to capture a valid exchange between the TLM instrument and the TLM simulator. The TLM Instrument receives a malicious payload (see Figure 28), which is then executed and opens a reverse shell to the attacker’s host (see Figure 29).

Figure 26: Exploit execution.

Figure 27: Capturing a valid TM Packet.

Figure 28: TLM Instrument receives a malicious packet.

Figure 29: Reverse shell from a TLM Instrument to the attacker's host.

**Recommendations**

In the exploitation scenario described above, we see two following vulnerabilities:

- ZMQ communication is unencrypted.
- Use of Pickle

To prevent the RCE, we recommend resolving both issues. Although replacing the plain ZMQ communication with ZMQ SSH Tunnelling might be tempting, more is needed. It will mitigate the MitM attacks; however, given that the TLM instrument opens a port and connects to a telemetry source without any verification, another attack vector emerges for exploitation – in case the bad actor can access the telemetry source host, they can stop a telemetry source and start their own with a malicious payload.

nttoole:
Addresses: 1) network security, 2) configuration file security. 

Fixes #531


EmilyPascua:
@nttoole - I'm merging this old PR in since it's relevant!

EmilyPascua (dismissed):

Issue #543, opened 2025-02-25
`ait-seq-encode` and `ait-seq-decode` are broken as they reference old method names


vatsal-j:
See https://github.com/NASA-AMMOS/AIT-Core/blob/70d6ed4f07107253d9a36ad799a05b1f5df7d42f/ait/core/bin/ait_seq_decode.py#L82 and
https://github.com/NASA-AMMOS/AIT-Core/blob/70d6ed4f07107253d9a36ad799a05b1f5df7d42f/ait/core/bin/ait_seq_encode.py#L88

which should be https://github.com/NASA-AMMOS/AIT-Core/blob/70d6ed4f07107253d9a36ad799a05b1f5df7d42f/ait/core/seq.py#L284 and https://github.com/NASA-AMMOS/AIT-Core/blob/70d6ed4f07107253d9a36ad799a05b1f5df7d42f/ait/core/seq.py#L250

nttoole:
Resolved by https://github.com/NASA-AMMOS/AIT-Core/pull/544

PR #544, opened 2025-02-25
Issue #543 - ait-seq-encode and ait-seq-decode reference old method names

vatsal-j:
Related to #504 which seemingly got overlooked. Perhaps a test for encoding and decoding sequences should be added?


MJJoyce (approved):
The 3.10 lint / test break isn't related to anything in this PR. Looks like a `black` error in an unrelated file. Can / should be fixed in a separate ticket.

+1 from me

nttoole (approved):

Issue #545, opened 2025-03-05
Free the dependencies!


eguzman-jpl:
There are a number of dependencies locked in the projects dependency specification. Are there any known incompatibilities with newer versions? This constraint prevents any downstream projects from receiving security or bug fixes. AIT-Core itself also becomes non-installable on projects where another version of said dependency is already specified. 

I am proposing to free (">=") all the dependencies.

PR #549, opened 2025-11-19
Free the dependencies!

eguzman-jpl:
Creating a PR for #545. 

There are a number of dependencies locked in the projects dependency specification. The current constraint on Python and other dependencies prevents any downstream projects from receiving security or bug fixes. AIT-Core itself also becomes non-installable on projects where another version of said dependency is already specified.

Current Stable Python version is 3.14.


nttoole:
Resolved by #553

Issue #557, opened 2026-08-25
Fix failing readthedocs + Update to latest config syntax


EmilyPascua:
ReadTheDocs [builds](https://app.readthedocs.org/projects/ait-core/) have been failing for sometime. Should be updated to use the latest syntax as defined [here](https://docs.readthedocs.com/platform/stable/config-file/index.html).

EmilyPascua:
docs are now passing! 🎊

PR #558, opened 2026-08-25
Issue #557 - Updating readthedocs.yml to use v2 syntax

EmilyPascua:
#557 - readthedocs builds have been failing for a while to do old config syntax. Updating to use v2, which is currently supported.


EmilyPascua:
Also needed to get #531 merged in.

nttoole (approved):

Issue #559, opened 2026-08-26
Config caching with msgpack migration


EmilyPascua:
As part of the pickle security refactoring to use msgpack in place of pickle for serialization, config caching was removed which in-turn will cause performance regressions. Initial PR for pickle removal: [here](https://github.com/NASA-AMMOS/AIT-Core/commit/dc9280b7b6eaa64c3275497166a6e31d360ce52e#diff-3355866b5750a9772c0d2d7d98c32b3e515c8a6dd227231f3f9db63908486176), note: `check_yaml_timestamps` was updated per the refactoring, however, `dirty` cache method was removed and therefore the `check_yaml_timestamps` method is no longer called. 

I spoke to the previous developer (@nttoole) who mentioned this was done intentionally as there was documented issues with the caching implementation which some missions have tried to improve in their own adaptations. In the long term and outside of the scope of this issue, it would probably be best to reevaluate and improve caching.

PR #560, opened 2026-08-26
issue #559 - adding back caching mechanism for msgpack

EmilyPascua:


nttoole:
Has the caching been tested (create cache, update dict file mod time, reload) ?

EmilyPascua:
@nttoole - PR still in progress for now. Just adding the changes that weren't migrated over. Not too certain how in depth we want to get into this since the real/long term change would be to to re-evaluate how the caching mechanism is done. Keeping this PR opened for now as an option.

nttoole:
Okey doke, perhaps it should be 'Convert to draft'ed until its ready

EmilyPascua:
> Okey doke, perhaps it should be 'Convert to draft'ed until its ready

Good idea, Nicko! I forgot you can do this -- I converted it! Thanks 👍🏼

PR #563, opened 2026-09-10
Create makefile for ease developer set-up

EmilyPascua:
Quick PR to add a Makefile and update README. Simplify the set-up steps.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,83 @@
+.PHONY: help dev-install install-python pre-commit setup-env setup test lint docs clean tox run-tests
+
+help:
+	@echo "AIT-Core Development Makefile"
+	@echo "============================="
+	@echo ""
+	@echo "Available targets:"
+	@echo "  setup-env      - Create and activate poetry virtual environment"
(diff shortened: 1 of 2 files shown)


EmilyPascua:
I didn't think an issue was relevant here because it's more-of a QOL thing and developer specific, but let me know if one is needed.

Makefile
```diff
@@ -0,0 +1,83 @@
...
+
+dev-install:
+	@echo "Installing package with development dependencies..."
+	poetry install
+
+install-python:
+	@echo "Installing required Python versions via pyenv..."
+	cat .python-version | xargs -I{} pyenv install --skip-existing {}
+
+pre-commit:
```
nttoole, line 38: Forcing the pre-commit setup eh?  😢 Prolly for the best tho...

EmilyPascua: Yes, indeed! 😈

EmilyPascua: Developers will just need to run `make setup` to be up in running. 👍🏼

